PR #2, opened 2017-08-23
More Python3 compatibility

jpgrayson:
This patch series takes a few more aggressive steps toward Python 3 compatibility. These changes necessarily break compatibility with Python 2.4 and 2.5 thus making the new minimum version Python 2.6.

The ultimate goal is to support Python 3.4+ and Python 2.7 out of the same codebase.


ctmarinas:
Looks great. Merged. Thanks.

PR #6, opened 2017-10-07
Python3 unicode

jpgrayson:
This patch series fully enables Python3 support for stgit.

Previous PR's addressed the relatively easy Python3 compatibility issues. This series addresses the big giant remaining issue: string handling. There are several different known, viable approaches to achieving Python 2+3 string compatibility. After experimenting with several, I finally landed on the `unicode_literals` approach. I.e. adding `from __future__ import unicode_literals` to force all string literals to be of type `unicode` in Python 2 which aligns them to Python 3 `str`. This drives the code to be more Python3-centric in its string vs. bytes hygiene. The key downside to this approach is that it forces this hygiene to be established in a non-incremental manner...but that's what this PR delivers.

The net of the changes in this series is that stgit works for Python 2.7 (including PyPy) as well as Python 3.3 through 3.6. All existing unit tests pass on these Python platforms.

Note that Python 2.6 _is not_ fully working. It may still be possible to resurrect support for Python 2.6, but given that Python 2.7 has been available for 7 years, my intent is to abandon Python 2.6.

Another note: there are at least a few use cases not captured by the test suite that can cause trouble. Specifically, cases where the locale is configured to C/POSIX and non-ascii strings enter the system via environment variables (e.g. GIT_AUTHOR_NAME) or the command line. I do not know of any issues when the user has a UTF8 locale configured (e.g. en_US.UTF-8).


ctmarinas:
Thanks.

PR #8, opened 2017-11-14
refresh: do not include submodules by default

jacob-keller:
When refreshing patches, the default action is to include every path
which is changed relative to the current working directory. This common
workflow creates problems when one uses a submodule. Due to the nature
of submodules, under some workflows they can often be out-of-sync, such
as when working on new updates within a submodule.

Accidental refreshes which include the submodule changes often occur,
which leads to commits which unintentionally include changes to
submodule pointers. These commits likely cause submodules to point to
incorrect locations.

Since it is not easy to convince many developers to always use "refresh
--index", instead, modify the default refresh behavior.

Instead of including all files, now check to make sure that we exclude
submodule files.

We can do this by using git ls-tree to list the files and their modes.
To avoid re-running ls-tree many times, we'll use a compiled regex to
match lines that look like submodules, and extract their filename.

Add a new option to refresh "--submodules" which tells refresh to
include submodules regardless. Additionally, add a "--no-submodules" and
a stg.refreshsubmodules configuration so that users may opt to configure
this behavior.

Since the number of users who use submodules is likely small, make this
the new default behavior.

Signed-off-by: Jacob Keller <jacob.e.keller@intel.com>

### DIFF
--- a/stgit/commands/pull.py
+++ b/stgit/commands/pull.py
@@ -114,7 +114,7 @@ def func(parser, options, args):
     post_rebase(crt_series, applied, options.nopush, options.merged)
 
     # maybe tidy up
-    if config.get('stgit.keepoptimized') == 'yes':
+    if config.getbool('stgit.keepoptimized'):
         git.repack()
 
     print_crt_patch(crt_series)
--- a/stgit/commands/refresh.py
+++ b/stgit/commands/refresh.py
@@ -6,6 +6,7 @@ from stgit import argparse, utils
 from stgit.argparse import opt
 from stgit.commands import common
 from stgit.lib import git, transaction, edit
+from stgit.config import config, GitConfigException
 from stgit.out import out
 
 __copyright__ = """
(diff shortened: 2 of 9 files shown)


jacob-keller:
This now also includes a separate patch to help avoid conflicts with submodules due to how update-index --refresh behaves.

jacob-keller:
I added a couple of test cases for these new features.

ctmarinas:
Since I don't normally use git submodules, I'd have to trust you the patches do the right thing ;). Anyway, following Peter's comment on the list, would you mind reposting an updated version (on the list) before a new pull request?

Thanks.

jacob-keller:
I've pushed the latest version to the pull request, but I'll re-send
patches as soon as I can (next Monday when I'm back in the office)

Thanks,
Jake

jacob-keller:
> Since I don't normally use git submodules, I'd have to trust you the patches do the right thing ;). Anyway, following Peter's comment on the list, would you mind reposting an updated version (on the list) before a new pull request?

I've sent an updated mail with all the current patches in this series. (I've rebased and put the patches in order, since they touch conflicting code)

PR #10, opened 2017-11-28
Repair import from mail without charset

jpgrayson:
Patches generated with `git format-patch` do not specify a charset, but
`stg import` was assuming that a charset would always be present. The
regression is repaired by assuming UTF-8 encoding when charset is not
specified in email payloads.

Signed-off-by: Peter Grayson <jpgrayson@gmail.com>


stgit/commands/common.py
```diff
@@ -444,6 +444,8 @@ def __decode_header(header):
                                        'application/octet-stream']:
             payload = part.get_payload(decode=True)
             charset = part.get_content_charset()
```
ctmarinas: get_content_charset('utf-8') should do, without the if block below.

ctmarinas:
While you are at it, do you have any idea why "stg edit" fails if a commit message as non-ascii characters (e.g. trying to add "Cc: David Kågedal <david@kagedal.org>")? "git commit --annotate" handles this properly but doing "stg uncommit && stg edit" fails. Thanks.

jpgrayson:
I believe I was able to reproduce the issue with `stg edit`. The issue I reproduced affected Python 2, but not Python 3. The latest PR adds a unit test that reproduces the issue along with a repair.

Sorry for the regression. Getting stgit to be unicode-clean is hard (at least for me). I doubt this will be the last issue. Please let me know if you see any more issues and I'll do my best to repair.

ctmarinas:
`stg mail -m` where the commit has non-ascii characters fails with similar error. Is there a way to force 'utf-8' instead of 'ascii' as the default decoder/encoder for python2? Alternatively, shall we drop support for python2 entirely? I would ask the latter on the stgit mailing list for wider visibility.

jpgrayson:
I am working on this problems. Again, thank you for helping identify these failing cases.

The approach I am currently taking is to modify `GIT_AUTHOR_NAME` and `GIT_COMMITTER_NAME` environment variables in`test-lib.sh` to have non-ASCII characters. This triggers several test failures for both Python 2 and 3; including `edit` and `mail` failures that seem to match the signature you are seeing.

I've made several repairs already to ensure that all environment variables ingress as unicode text. That repairs the bulk of the issues, but there remain several other issues that will take me more time to address.

Regarding supporing both Python 2 and 3, I agree that it is tempting to drop support for Python 2. My goal has been to have a stgit release that supports both 2 and 3. I expect that such a stgit release will aid at least some users' migration to Python 3. It's definitely challenging dealing with both 2 and 3 at the same time, but I also still believe it is possible to iron-out the unicode rough edges without abandoning Python 2.

Once we make a good/stable stgit release supporting Python 2 and 3, then I would be in favor of dropping Python 2 support.

I will also note that w.r.t. unicode difficulties, it is not just Python 2 vs 3 that makes it hard to make stgit unicode-clean; it is also that the test suite runs with `LC_ALL=C`. This means that even in a pure Python 3 environment, we have to force a UTF-8 encoding at all the locale-aware OS interfaces: paths, files, environment variables, command line arguments, etc. I think it is a nice-to-have feature for stgit to Just Work even in environments with misconfigured locales, but it would be a stgit code simplifier to put the onus on users to have a UTF-8 locale configured--especially if coupled with Python 3-only support.

jpgrayson:
The latest set of patches much more comprehensively addresses the various issues arising from non-ascii characters being in git author/committer names and environment variables.

ctmarinas:
Merged now, thanks. Sorry for the delay.

PR #16, opened 2018-01-28
Handle utf-8 email bodies with latin-1 characters

jpgrayson:
It is possible for emails to specify a utf-8 charset, but erroneously
contain latin-1 encoded characters. Reference:

  https://lkml.org/lkml/2018/1/24/704

Instead of failing import with a decoding error, stgit now takes the same
approach as git by decoding non-utf8 bytes as latin-1.

Signed-off-by: Peter Grayson <jpgrayson@gmail.com>


ctmarinas:
Thanks for sorting this out.

PR #29, opened 2019-03-01
Add --overwrite option for publish.

andrewchambers:
This option is useful for 'github style'
pull requests where it is often requested
to rewrite your pull request multiple
times before it is accepted.


jpgrayson:
Thank you @andrewchambers for this PR. Although `stg publish` is not in my normal workflows, I think having this new `--overwrite` option makes sense. The code looks good to me as well. Merged.

PR #33, opened 2019-06-21
Add 'domain' parameter to 'mail' command

tohojo:
The default message ID generation in the Python email library will use the
system hostname as the domain part. This adds an option to override
this (for Python versions that support it), to avoid leaking the system
hostname into the sent messages.

### DIFF
--- a/completion/stgit.zsh
+++ b/completion/stgit.zsh
@@ -265,6 +265,7 @@ _stg-mail() {
         '(-E --edit-patches)'{-E,--edit-patches}'[edit patches before sending]'
         '(-s --sleep)'{-s,--sleep=}'[seconds to sleep between sending emails]:seconds'
         '--in-reply-to=[reply reference id]:refid'
+        '--domain=[domain to use for message ID]:domain'
         '(-u --user)'{-u,--user=}'[username for SMTP authentication]:user'
         '(-p --password)'{-p,--password=}'[password for SMTP authentication]:password'
         '(-T --smtp-tls)'{-t,--smtp-tls}'[use TLS for SMTP authentication]'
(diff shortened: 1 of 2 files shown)


jpgrayson:
Hi @tohojo. Thank you for this PR! Adding this `--domain` option makes sense to me.

That said, I would prefer we handle the Python compatibility issue a little differently. When running with a version of Python that does not support the `domain` argument to `email.util.make_msgid()`, I would prefer that we explicitly fail (i.e. `raise CmdException(...)`) rather than silently ignoring the user's `--domain` option. I suppose a simple test of `sys.version_info < (3, 2)` would be sufficient to determine whether the feature is supported.

Also, please make sure `make lint` succeeds. It appears there may be a too long line.

For bonus points, consider adding this option to `completions/stgit.zsh`. The fish and bash completions are autogenerated, but the zshell completions are manually curated.

Thanks again!

tohojo:
Pete Grayson <notifications@github.com> writes:

> Hi @tohojo. Thank you for this PR! Adding this `--domain` option makes
> sense to me.

Cool!

> That said, I would prefer we handle the Python compatibility issue a
> little differently. When running with a version of Python that does
> not support the `domain` argument to `email.util.make_msgid()`, I
> would prefer that we explicitly fail (i.e. `raise CmdException(...)`)
> rather than silently ignoring the user's `--domain` option. I suppose
> a simple test of `sys.version_info < (3, 2)` would be sufficient to
> determine whether the feature is supported.

Sure, sounds reasonable, will fix.

> Also, please make sure `make lint` succeeds. It appears there may be a
> too long line.

Will fix this as well.

> For bonus points, consider adding this option to
> `completions/stgit.zsh`. The fish and bash completions are
> autogenerated, but the zshell completions are manually curated.

Ah. I looked at the bash completions, saw they were automatically
generated, and didn't look any further. I can certainly update the zsh
completions as well (I'm as zsh user myself) :)

jpgrayson:
Merged. Thank you!

Issue #37, opened 2019-08-15
setup.py installation broken under Fedora30


jnsnow:
Running:

```
> python3 setup.py install --user
```

appears to produce a broken installation:

```
jhuston@probe (master) ~/s/stgit> which stg
/home/bos/jhuston/.local/bin/stg
jhuston@probe (master) ~/s/stgit> stg --help
Traceback (most recent call last):
  File "/home/bos/jhuston/.local/bin/stg", line 51, in <module>
    from stgit.main import main
  File "/home/bos/jhuston/.local/lib/python3.7/site-packages/stgit/main.py", line 15, in <module>
    from stgit import argparse, run, utils
  File "/home/bos/jhuston/.local/lib/python3.7/site-packages/stgit/argparse.py", line 20, in <module>
    from stgit.lib.git import Date
ModuleNotFoundError: No module named 'stgit.lib.git'
```

It doesn't look like it installs anything under lib/git:

```
jhuston@probe (master) ~/s/stgit> ls ~/.local/lib/python3.7/site-packages/stgit/lib
edit.py  __init__.py  log.py  objcache.py  __pycache__/  stack.py  stackupgrade.py  transaction.py
```

Editing the setup.py configuration appears to fix it:

```
diff --git a/setup.py b/setup.py
index 264a58b..1cc8936 100755
--- a/setup.py
+++ b/setup.py
@@ -110,7 +110,10 @@ setup(
     description='Stacked GIT',
     long_description='Push/pop utility on top of GIT',
     scripts=['stg'],
-    packages=list(map(str, ['stgit', 'stgit.commands', 'stgit.lib'])),
+    packages=list(map(str, ['stgit',
+                            'stgit.commands',
+                            'stgit.lib',
+                            'stgit.lib.git'])),
     data_files=[
         ('share/stgit/templates', glob('stgit/templates/*.tmpl')),
         ('share/stgit/examples', glob('examples/*.tmpl')),
```

lkraav:
I can confirm, same thing happens here.

jpgrayson:
Repaired in #38 (d0dcf0ee3466f869d122a1044c9d25cfc1954bc9).

PR #38, opened 2019-08-15
setup.py: include lib.git submodule

jnsnow:
stgit.lib.git doesn't appear to be included otherwise.

While we're here, delete a list(map(str, [...])) invocation that
seemingly does nothing.

Might be an appropriate fix for issue #37.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -110,7 +110,8 @@ setup(
     description='Stacked GIT',
     long_description='Push/pop utility on top of GIT',
     scripts=['stg'],
-    packages=list(map(str, ['stgit', 'stgit.commands', 'stgit.lib'])),
+    packages=list(map(str, ['stgit', 'stgit.commands',
+                            'stgit.lib', 'stgit.lib.git'])),
     data_files=[
         ('share/stgit/templates', glob('stgit/templates/*.tmpl')),
         ('share/stgit/examples', glob('examples/*.tmpl')),


jpgrayson:
Hi @jnsnow. Thanks for this patch. The `stgit.lib.git` is an omission--good catch.

However, the `list(map(str, [...]))` madness actually has a purpose. Because we are using `from __future__ import unicode_literals`, the type of the items in this `packages` list is `str` for Python 3, but `unicode` for Python 2. These package strings need to be `str` for Python 2 as well (for "reasons"), thus the `list(map(str, [...]))` business converts from `unicode` to `str` in a Python 2 environment, but is idempotent in a Python 3 environment.

The [failing travis builds](https://travis-ci.org/ctmarinas/stgit/builds/572485397) are due to this issue.

We could potentially solve this by removing `unicode_literals` from `setup.py`, but currently every source file in stgit uses the same set of `__future__` imports so that we have the same Python 2/3 compatibility concerns and solutions in every file.

jnsnow:
Oh, I see! I was wondering what I blew up. I'll respin. Thanks!

lkraav:
Just adding `stgit.lib.git` to `map()` seems to work.

```
leho@gusto /mnt/datapool/src/stgit.git $ stg --version
Stacked GIT 0.19-80-g114b-dirty                       
git version 2.21.0                                    
Python version 3.6.5 (default, Jul 21 2019, 00:56:23) 
[GCC 8.3.0]
```

jpgrayson:
Rebased to ctmarinas:master. Thanks!

PR #40, opened 2019-08-30
out.py: prefer stderr and remove isatty() restriction

### LINKED ISSUE #35: cannot redirect output
When using `stg import -M path/to/mail.mbox`, the output cannot be redirected to file:

jhuston@probe (review) ~/s/qemu> stg import -M etc.mbox > output.txt
jhuston@probe (review) ~/s/qemu> cat output.txt 
jhuston@probe (review) ~/s/qemu> 

The file does have valid patches in it, though:

jhuston@probe (review) ~/s/qemu> stg import -M etc.mbox
Checking for changes in the working directory ... done
Importing patch "iotests-fix-_filter_img_create" ... done
Importing patch "vmdk-use-bdrv_dirname-for" ... done
Importing patch "iotests-keep-testing-broken" ... done
Importing patch "vmdk-reject-invalid-compressed" ... done
Importing patch "iotests-disable-broken" ... done
Importing patch "iotests-disable-110-for-vmdk" ... done
Importing patch "iotests-disable-126-for-flat" ... done
Now at patch "iotests-disable-126-for-flat"

And the command will succeed with redirected output, but the output isn't displayed on screen OR in the file.

I guess this is from: from stgit.out import out, which is a MessagePrinter();
which makes this look unfortunately intentional:

	if file or sys.stdout.isatty():
            self.__out = self.__stdout
        else:
            self.__out = Output(None)

The reason I would like this to work is to be able to script using stgit in other tools, and the same isatty check prevents me from getting output via e.g. python's subprocess call or other scripts.

jnsnow:
Maybe this is good enough for now to close #35.

By changing .info() and .note() to stderr, we can slacken the isatty() restriction and get away with no test breakages. Seems fairly clean; though of course this is still a change in behavior that might break fragile scripts.

### DIFF
--- a/stgit/out.py
+++ b/stgit/out.py
@@ -99,10 +99,7 @@ class MessagePrinter(object):
                     sys.stderr.fileno(), 'w', buffering=1, encoding='utf-8'
                 )
             )
-        if file or sys.stdout.isatty():
-            self.__out = self.__stdout
-        else:
-            self.__out = Output(None)
+        self.__out = self.__stdout
 
     def stdout(self, line):
         """Write a line to stdout."""
--- a/stgit/out.py
+++ b/stgit/out.py
@@ -127,10 +127,10 @@ class MessagePrinter(object):
 
     def info(self, *msgs):
         for msg in msgs:
-            self.__out.single_line(msg)
+            self.__stderr.single_line(msg)
 
     def note(self, *msgs, **kw):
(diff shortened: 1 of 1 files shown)


jpgrayson:
I think this looks great. Thanks for making it happen.

I'd like to give @ctmarinas a chance to veto this if there is some historical context we might be missing. Otherwise, I'll merge this in the next couple days.

PR #43, opened 2019-10-09
Make the build reproducible

lamby:
Whilst working on the [Reproducible Builds effort](https://reproducible-builds.org/) we noticed that stgit could not be built reproducibly.

This is because the contents of the dynamically-compiled Bash completion script was not being generated in a deterministic manner, as well as the `cmdlist.py` module.

This was originally filed in Debian as [#942009](https://bugs.debian.org/942009).

### DIFF
--- a/stgit/commands/__init__.py
+++ b/stgit/commands/__init__.py
@@ -73,7 +73,7 @@ def get_commands(allow_cached=True):
 def py_commands(commands, f):
     f.write('from __future__ import unicode_literals\n\n')
     f.write('command_list = {\n')
-    for name, (mod, kind, help) in commands.items():
+    for name, (mod, kind, help) in sorted(commands.items()):
         f.write('    %r: (\n' % name)
         f.write('        %r,\n' % mod)
         f.write('        %r,\n' % kind)
--- a/stgit/completion/bash.py
+++ b/stgit/completion/bash.py
@@ -26,7 +26,7 @@ class CompgenBase(object):
             cmd += ['-A', act]
         words = self.words(var)
         if words:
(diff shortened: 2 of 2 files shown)


jpgrayson:
Hi @lamby. Thank you for identifying this reproducible build issue and thank you for this PR.

As I looked at this patch and stgit as a whole, I saw opportunities to solve the problem in a different, deeper manner. I also identified reproducibility issues in the generated `stg.fish` completions that were not addressed by this PR. The fish problems may only have affected Python < 3.6.

In net, I have pushed a more comprehensive repair for this issue, so I am closing this PR.

Questions: is there a way to get feedback from the reproducible builds project about whether stgit now passes its tests? Or will we have to release stgit 0.21(0.20 was released recently) and wait for it to make it into debian unstable?

lamby:
Hey, thanks for this!

> Questions: is there a way to get feedback from the reproducible builds project about whether stgit now passes its tests? Or will we have to release stgit 0.21(0.20 was released recently) and wait for it to make it into debian unstable?

From the viewpoint of our own testing framework, yes, we will have to wait for the relevant version to be updated in Debian unstable, alas.

bmwiedemann:
It seems the fix was commit 883f8a0ca67b4b16642bfce011023456ce11a9fe 
and I was able to reproduce the unreproducibilty of plain 0.20
and adding the patch ontop made it reproducible in the openSUSE reproducible-builds context. We are not varying all things Debian does, so further issues could be found there later, but it is a good sign.

jpgrayson:
@bmwiedemann, thank you for this feedback. This provides some motivation to cut a stgit 0.21 release soon, so that we will have a reproducible (hopefully) version of stgit out there.

PR #46, opened 2019-12-18
Fix importing mail with CRLF rather than LF

gwd:
I've been having trouble importing some patch series due to CRLF incompatibilities.  This strips the CR from incoming mail by default before processing, unless `--keepcr` is specified (similar to the behavior of `git am`).

It also includes a fix to the development branch where `stg import` always edits every commit even though `--edit` wasn't specified.

### DIFF
--- a/stgit/commands/common.py
+++ b/stgit/commands/common.py
@@ -561,11 +561,14 @@ def update_commit_data(cd, options):
             add_sign_line(cd.message, sign_str,
                           cd.committer.name, cd.committer.email))
 
-    # Let user edit the commit message manually, unless
-    # --save-template or --message was specified.
+    # Edit the commit message manually if:
+    # - A message was specified, but --edit was also specified.
+    # - No message was specified, and --save_template was also not specified
     if (
-        not getattr(options, 'save_template', None)
-        and getattr(options, 'message', None) is None
+        (getattr(cd, 'message', None)
+         and getattr(options, 'edit', None))
+        or (not getattr(cd, 'message', None)
+            and not getattr(options, 'save_template', None))
     ):
         tmpl = templates.get_template('patchdescr.tmpl')
         if tmpl:
--- a/stgit/commands/common.py
+++ b/stgit/commands/common.py
@@ -428,7 +428,7 @@ def __parse_description(descr):
     return (subject + body, authname, authemail, authdate)
 
 
-def parse_mail(msg):
+def parse_mail(msg, options):
     """Parse the message object and return (description, authname,
     authemail, authdate, diff)
     """
@@ -482,6 +482,14 @@ def parse_mail(msg):
                                        'application/octet-stream']:
             payload = part.get_payload(decode=True)
             msg_data += payload
+
+    # RFC-compliant mail uses CRLF.  Replace this with LF unless
+    # otherwise instructed.
+    if not getattr(options, 'keepcr', False):
+        msg_data_lf = b''
+        for line in msg_data.splitlines():
+            msg_data_lf += line + b'\n'
+        msg_data = msg_data_lf
 
     rem_descr, diff = __split_descr_diff(msg_data)
     if rem_descr:
--- a/stgit/commands/imprt.py
+++ b/stgit/commands/imprt.py
@@ -149,6 +149,11 @@ options = [
         action='store_true',
         short='Show the patch content in the editor buffer',
     ),
+    opt(
+        '--keepcr',
+        action='store_true',
+        short='Do not strip carriage returns when importing mail messages',
+    ),
 ] + argparse.author_options() + argparse.sign_options()
 
 directory = DirectoryHasRepository()
@@ -289,7 +294,7 @@ def __import_file(filename, options, patch=None):
             raise CmdException('error parsing the e-mail file: %s' % str(ex))
         (
             message, author_name, author_email, author_date, diff
-        ) = parse_mail(msg)
+        ) = parse_mail(msg, options)
     else:
         patch_str = f.read()
         (
@@ -370,7 +375,7 @@ def __import_mbox(filename, options):
                  author_name,
                  author_email,
                  author_date,
-                 diff) = parse_mail(msg)
+                 diff) = parse_mail(msg, options)
                 __create_patch(None, message, author_name, author_email,
                                author_date, diff, options)
     finally:


stgit/commands/common.py
```diff
@@ -483,6 +483,14 @@ def __decode_header(header):
             payload = part.get_payload(decode=True)
             msg_data += payload
 
+    # RFC-compliant mail uses CRLF.  Replace this with LF unless
+    # otherwise instructed.
+    if not getattr(options, 'keepcr', False):
+        msg_data_lf = b''
+        for line in msg_data.splitlines():
+            msg_data_lf += line + b'\n'
+        msg_data = msg_data_lf
```
jpgrayson: Why not `msg_data = msg_data.replace(b'\r\n', b'\n')`?

stgit/commands/common.py
```diff
@@ -561,11 +569,14 @@ def update_commit_data(cd, options):
             add_sign_line(cd.message, sign_str,
                           cd.committer.name, cd.committer.email))
 
-    # Let user edit the commit message manually, unless
-    # --save-template or --message was specified.
+    # Edit the commit message manually if:
+    # - A message was specified, but --edit was also specified.
+    # - No message was specified, and --save_template was also not specified
```
jpgrayson: Spelling: `--save-template`.

stgit/commands/common.py
```diff
@@ -561,11 +569,14 @@ def update_commit_data(cd, options):
...
     if (
-        not getattr(options, 'save_template', None)
-        and getattr(options, 'message', None) is None
+        (getattr(cd, 'message', None)
+         and getattr(options, 'edit', None))
+        or (not getattr(cd, 'message', None)
+            and not getattr(options, 'save_template', None))
```
jpgrayson: Using this `update_commit_data()` function in `imprt.py` is certainly proving to be problematic.

Consider the following alternative predicate:

```
if getattr(options, 'edit', None) or (
    not getattr(options, 'save_template', False)
    and getattr(options, 'message', None) is None
    and getattr(options, 'edit', None) is None
):
    ...
```

`getattr(options, 'message', None)` will only be true if the user provided a `-m/--message` option on the command line whereas `cd.message` will be true if there is a non-empty message associated with the commit. For purposes of this predicate, I don't think we want to consider the commit message itself, only whether the user supplied `-m/--message`. Similarly `getattr(options, 'edit', None)` will only be true if the user provided a `--edit` command line option.

In the long run, I'm thinking that `update_commit_data()` should perhaps take explicit arguments for `save_template`, `edit`, and `user_message` instead of inspecting `options` which is wildly different depending on the particular command.

stgit/commands/common.py
```diff
@@ -428,7 +428,7 @@ def __parse_description(descr):
     return (subject + body, authname, authemail, authdate)
 
 
-def parse_mail(msg):
+def parse_mail(msg, options):
```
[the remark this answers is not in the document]
gwd: That was my initial implementation actually.  But `parse_mail` is only called in two places, both in `imprt.py`, and both places ended up looking like this:

```python
    parse_mail(msg, getattr(options, 'keepcr', False))
``` 

Having a single place where we defined how `keepcr` behaved seemed to be more inline with DRY.  And if there were ever more options you wanted to add to affect mail parsing, you wouldn't have to plumb them through, just add them to the options list and look at them in `parse_mail`.

stgit/commands/common.py
gwd: As you can probably tell, python isn't my main language. :smile: This is adapted from a code snipped on Stackoverflow.

I had pondered why the commenter had used `splitlines` rather than `replace`.  One potential advantage is that the built-in `splitlines` should automatically handle *any* alternate line endings, not just CRLF.  It could also potentially be optimized due to the fact that it's not a general-purpose string replacement.

I don't know if either of those will turn out to be real advantages in practice, but in general, using a core library to do the thing you want rather than implementing it yourself seems a good idea, so I kept it that way.

stgit/commands/common.py
```diff
@@ -561,11 +569,14 @@ def update_commit_data(cd, options):
...
 
-    # Let user edit the commit message manually, unless
-    # --save-template or --message was specified.
+    # Edit the commit message manually if:
+    # - A message was specified, but --edit was also specified.
+    # - No message was specified, and --save_template was also not specified
     if (
-        not getattr(options, 'save_template', None)
-        and getattr(options, 'message', None) is None
+        (getattr(cd, 'message', None)
```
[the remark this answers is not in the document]
gwd: Ack

stgit/commands/common.py
gwd: > For purposes of this predicate, I don't think we want to consider the commit message itself, only whether the user supplied `-m/--message`.

But that's basically the whole problem with using this for `import`, isn't it?  The general case is that people *don't* want to edit the commit message; rather, they want to use the one that's already there.  The truth table looks like this:

| command | --message | --edit | --save-template | Edit? |
|-|-|-|-|-|
| new / publish | N | n/a | N | Y |
| new / publish | Y | n/a | N | N |
| new / publish | N | n/a | Y | N |
| new / publish | Y | n/a | Y | N |
| import | n/a | N | n/a | N |
| import | n/a | Y | n/a | Y |

This is case where having `update_commit_data()` take an `invoke_editor` boolean might make more sense, and have the options logic decoded by the callers.

If you want I can refactor it that way.  OTOH, I only really wrote this patch so that the second patch would be usable for me.  I'm also happy to have you just take the `--keepcr` patch and fix the `--edit` issue the way you think best.

gwd: Oh, it should also be pointed out: `cd.message` will be set further up the function if there's a valid `options.message`.  So if `--message` is provided `cd.message` will always be non-`None`.  The only risk in the logic of my submission would be if `cd.message` was somehow set by `publish` or `new` when no `--message` was provided.

jpgrayson:
I have pushed your fix for the import editing problem along with some test updates to catch the erroneous import-time editor invocations. And I reworked `update_commit_data()` to take explicit keyword args instead of trolling through options.

So I think the `stg import` regression is resolved. Thank you for identifying and repairing the issue!

---

Regarding the CRLF mail import change, I've been working on that too, with your patch as a starting point. What I'm looking at doing is replacing the current email parsing, which uses the python stdlib `email` module, with `git mailsplit` and `git mailinfo`. I'm also still evaluating whether `git am` could be used directly.

This path would more closely align stgit's behavior with git's behavior, and will specifically allow the `--keep-cr` option to be passed-through to either `git mailsplit` or `git am`.

gwd:
I think in general reusing git's implementation of this thing rather than doing a re-implementation in python is a good idea: any future improvements you get to pick up "automatically" rather than waiting for people to trip over them and them fix them yourself.

Unfortunately, git's handling of this is broken at the moment for base64-encoded mail.  The CRLF replacement is done in `mailsplit`, but `mailsplit` happens before the base64 decoding; so if you try to `git am` a base64-encoded patch, it still fails.  I've reported the issue here: https://public-inbox.org/git/c44c3958-b0eb-22bd-bc35-04982706162f@citrix.com/ but haven't received much response yet.

So, if you do switch to using `git mailsplit` and `git mailinfo`, it will be broken for base64-encoded emails until that's fixed upsteam, and then that fix trickles down to all current versions of git.

jpgrayson:
I have pushed some changes that I believe resolve this issue satisfactorily.

- New tests for email files and mbox files with CRLF line endings
- Use `git mailsplit` and `git mailinfo` instead of python's `email` and `mailbox` modules for mail import
- Add `--keep-cr` option to `stg import`
- Add tests for `--keep-cr`.

I note that the change proposed in this PR was not actually effective, at least on platforms with native LF line endings. The Python email libraries strip CR line endings by default even though this is not strictly RFC-compliant behavior.

https://docs.python.org/3/library/email.policy.html#email.policy.default

This PR assumed that line endings coming out of the python `email` processing would be unmodified from the source email file, but that proves not to be the case.

With the alternative approach of using `git mailsplit`, we inherit git's default behavior, which also strips CR characters, but we also get to piggyback on the `--keep-cr` option, which allows CR characters to be correctly preserved as sought by this PR.

N.B. the `stg import` option name is now spelled `--keep-cr` which matches the `git mailsplit` option name, but is different than the `--keepcr` naming used in this PR.

Thanks again @gwd for getting the ball rolling with this PR!

PR #50, opened 2020-02-01
Add mention of --cleanup and --delete in tutorial

jkleckner:
An important part of the workflow is cleaning up stgit information from branches that you no longer want to manage with stgit.

This adds mention of that in the tutorial.


jkleckner:
Seems like the coverage is counting comments and documentation...

jpgrayson:
Merged.

Thank you for this PR. I think this is a nice addition to the tutorial.

And sorry for the delay in addressing the PR.

PR #51, opened 2020-02-01
Clarify top for float command is current

jkleckner:
The `stg float` command will push a patch to a top which is the current patch which does not include unapplied patches.

This adds mentions that in the help text.

### DIFF
--- a/stgit/commands/float.py
+++ b/stgit/commands/float.py
@@ -40,10 +40,11 @@ kind = 'stack'
 usage = ['[--] <patches>',
          '-s <series>']
 description = """
-Push a patch or a range of patches to the top even if applied. The
-necessary pop and push operations will be performed to accomplish
-this. The '--series' option can be used to rearrange the (top) patches
-as specified by the given series file (or the standard input)."""
+Float a patch or range of patches to be the top-most applied patches.
+The patches to be floated may currently be either applied or unapplied.
+The necessary pop and push operations will be performed to float the
+named patches.  Patches not specified will remain applied or unapplied
+as they were prior to the float operation."""
 
 args = [patch_range('applied_patches', 'unapplied_patches')]
 options = [


jkleckner:
Seems like the coverage is counting comments and documentation...

stgit/commands/float.py
```diff
@@ -43,7 +43,9 @@
 Push a patch or a range of patches to the top even if applied. The
 necessary pop and push operations will be performed to accomplish
 this. The '--series' option can be used to rearrange the (top) patches
-as specified by the given series file (or the standard input)."""
+as specified by the given series file (or the standard input).
+Top is defined to be the currently selected patch set and, unless
+specified in a series, will not include unapplied patches."""
```
jpgrayson: There are a couple of definitional corrections needed, but even with those applied, I'm not sure this new sentence adds clarity. The new sentence both tries to define "top" and make an assertion about float's semantics. Two separate sentences would be better.

"currently selected" ==> "top-most applied"
"patch set" ==> "patch"

What do you think about this alternative description?:

> Float a patch or range of patches to be the top-most applied patches. The patches
> to be floated may currently be either applied or unapplied. The necessary pop and
> push operations will be performed to float the named patches. Patches not specified
> will remain applied or unapplied as they were prior to the float operation.

I tried to lay out the key definition that "floating" a patch is to make it the "top-most applied" patch, which covers one of the issues I think you're pointing out. I also added a sentence that tries to spell-out that float does not affect the applied/unapplied status of other patches. Lastly, I dropped the specific mention of the `--series` option. I don't think `--series` is useful enough to warrant being in the description.

jkleckner: Yes, I think that is more clear.  I will amend the patch with that text.

jkleckner: Tell me what you think of the short description "Move a patch to be the top applied patch".

jpgrayson: I think the previous verbiage is more on-target. The suggested short description indicates that only a single patch may be floated whereas the command allows for multiple patches which the existing short description indicates. Also, I think I prefer to use the existing "push" verb instead of "move" since `float` really is a lot like `push`, except that it works on already applied patches too.

jkleckner: Ok, reverted.

jpgrayson:
Merged.

I really appreciate you (@jkleckner) taking the time to help improve StGit's documentation. I know I tend to neglect the docs, so it's great to have someone giving them attention.

PR #59, opened 2020-05-04
Add config options for specifiing text added by 'pick --expose'

koct9i:
Config options "stgit.pick.prepend" and "stgit.pick.append" defines
format strings for 'git show --pretty'.

Option 'stgit.pick.prepend' inserts lines after commit subject.

Default is stgit.pick.append = 'format:(imported from commit %H)'.

Signed-off-by: Konstantin Khlebnikov <khlebnikov@yandex-team.ru>

### DIFF
--- a/stgit/commands/pick.py
+++ b/stgit/commands/pick.py
@@ -13,9 +13,11 @@ from stgit.commands.common import (
     parse_rev,
     print_current_patch,
 )
+from stgit.config import config
 from stgit.lib.git import CommitData, MergeConflictException, MergeException, Person
 from stgit.lib.transaction import StackTransaction, TransactionHalted
 from stgit.out import out
+from stgit.run import Run
 from stgit.utils import STGIT_CONFLICT, STGIT_SUCCESS, find_patch_name, make_patch_name
 
 __copyright__ = """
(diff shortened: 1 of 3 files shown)


jpgrayson:
Hi @koct9i. Thank you for this PR.

The general concept of this feature seems fine, but the whitespace handling regresses from `stg pick --expose`'s current behavior.

The current code ensures that there is a blank line between the original message body and the appended suffix, but the new code appends the suffix without an intervening newline.

Also with the new prepended prefix, the prepended lines are inserted after the subject line (which makes sense), but there is nothing that ensures that the conventional blank line after the subject line is maintained. Nor is there anything ensuring a blank line between the prepended prefix and the original message body.

For example, given a commit with this original message:

```
Subject0

Body2
Body3
```

Using `stgit.pick.prepend = "PREPEND"`, we will end up with:

```
Subject0
PREPEND

Body2
Body3
```

Whereas I would perhaps expect this:

```
Subject0

PREPEND

Body2
Body3
```

Given the flexibility of `--pretty` formats, I wonder if perhaps we might be better served by a single `stgit.pick.expose_format` configuration instead of separate `prepend` and `append` configuration. For example, I think this would produce `stg pick`'s current behavior:

`git show --no-patch --pretty="%B%n(imported from commit %H)"`

The above could be the basis for the default `stgit.pick.expose_format` and a user-defined format with a prefix might look something like:

`git config stgit.pick.expose_format '%s%n%nMY PREFIX%n%+b%nMY SUFFIX'`

Setting a non-default `stgit.pick.expose_format` seems like an advanced enough use-case that it seems okay to expose all the pretty format complexity (and power) to the user.

I have pushed a change to the test suite to more thoroughly check the current expose behavior w.r.t. newlines. It exposes the regression that this patch introduces.

I'll also note that adding a new test case or two for this feature would be really helpful for me to be able to merge the PR with confidence.

jpgrayson:
Hi @koct9i. Thank you for your thorough updates to this PR. I really appreciate having the new tests.

And apologies for taking so long to re-review.

Merged.

Thank you again!

PR #63, opened 2020-07-18
Add pre-commit hook in refresh

### LINKED ISSUE #45: `stg` workflow with commit hooks?
Packages like https://github.com/typicode/husky are making git hooks ~~great again~~ easy.

Example husky pre-commit hooks setup: https://github.com/web-padawan/aybolit/blob/master/package.json#L75

Example `package.json "lint-staged"` key: https://github.com/web-padawan/aybolit/blob/master/package.json#L80

I'm noticing neither `stg refresh` or  `stg commit` trigger these (at least?) pre-commit hooks. This creates risks of offroading commit quality and causes future PR difficulties, because `stg` users commits aren't being auto-corrected by `prettier --fix`, but native `git` users commits are. Their PRs now often also include involuntary corrections of previous `stg` commits, or require special attention to exclude hunks.

Any thoughts towards what a solution could be? Am I missing something simple about how I should configure `stgit` to work with commit hooks?

akshedu:
fix #45 

The aim of this PR is to add `pre-commit` support in `refresh` command. The hook will be triggered after `update-index` and if the hook fails (which usually means some files were updated) `refresh` will exit with a message to the effect. The user then has the option to review the changes and add them using `stg add` and then re run `refresh` to update the patch.


lkraav:
:+1: 

Were you able to take a look at adding any tests for this functionality?

lkraav:
Also, please add "fix #45" somewhere, so GH can make the connection?

akshedu:
> 👍
> 
> Were you able to take a look at adding any tests for this functionality?

Not yet. One of the tests is failing due to this change. I will check it out and also write tests for pre-commit hook

jpgrayson:
This looks like its on the right track. Thanks for taking on this feature.

Definitely need new tests and all tests passing before the PR will be accepted.

Note: I just pushed a change to support the latest isort 5.x. It covers some of the same ground as your 17e3d75 commit, but is more comprehensive. I suggest rebasing your commits.

akshedu:
> This looks like its on the right track. Thanks for taking on this feature.
> 
> Definitely need new tests and all tests passing before the PR will be accepted.
> 
> Note: I just pushed a change to support the latest isort 5.x. It covers some of the same ground as your [17e3d75](https://github.com/stacked-git/stgit/commit/17e3d752d5f9997e767b5f49d9dfb101115ab2ed) commit, but is more comprehensive. I suggest rebasing your commits.

Thanks @jpgrayson, I have rebased my branch with changes in master

akshedu:
Fixed the failing test, running `write-tree` as well with `update-index` before running the hook.

Will be adding tests for this feature soon.

akshedu:
@jpgrayson I have pushed tests, but somehow they are failing on github workflows while they are passing on my local. Do you think you can guide me what could be potentially wrong? Thanks!

jpgrayson:
@akshedu, I will try to find some time in the next few days to take a deeper look at the new test cases and the test failure(s).

akshedu:
Thanks @jpgrayson.

I figure out the reason for failing tests, I was running on my local shell which was `bash` in mac while the tests were running on Linux which was `sh` I think. I created a local docker image and checked the commands there and realised the `pre-commit` hook was not executing on Linux since the shell commands were not the same. I have fixed the issue and now all tests are passing.

Please review whenever you get time. And also if you think I need to add more test scenarios, let me know. Thanks,

t/t2703-refresh-pre-commit-hook.sh
```diff
@@ -0,0 +1,129 @@
...
+    command_error stg refresh 2>&1 |
+    grep -e "pre-commit hook failed" &&
+    git restore --staged file
+'
+
+test_expect_success 'refresh with path limiting, failing hook' '
+    echo "pre-commit-hook-path-limiting-fail" >> file &&
+    command_error stg refresh file 2>&1 |
+    grep -e "pre-commit hook failed" &&
+    git restore --staged file
```
jpgrayson, line 69: `git restore` is very new and remains experimental, first appearing in Git 2.23.0 (August 2019). StGit otherwise supports git versions back to 2.2.0 (November 2014).

I would like to avoid having to update StGit's Git compatibility and avoid compatibility problems on various platforms that might not have a very recent version of Git.

Please use `git reset` instead.

t/t2703-refresh-pre-commit-hook.sh
```diff
@@ -0,0 +1,129 @@
...
+    echo "pre-commit-hook-non-executable" >> file &&
+    stg refresh
+'
+
+# now a hook that edits the files added in index
+# The hook tests for trailing white spaces
+# If finds files, then fixes them and returns non-zero exit status
+test_expect_success 'install pre-commit hook' '
+    cp "$TEST_DIRECTORY"/t2703/pre-commit $HOOK &&
+    chmod +x "$HOOK"
```
jpgrayson, line 88: This script could be boiled-down such that it could reasonably be inlined here instead of existing as a separate file.

```sh
git diff-index --check HEAD -- && exit
sed -i 's/[[:space:]]*$//' file
exit 1
```

Not a bit deal, just something to consider.

stgit/commands/refresh.py
```diff
@@ -354,10 +354,21 @@ def func(parser, options, args):
...
+
+    # Run pre-commit hook, if fails, abort refresh
+    if not options.no_verify:
+        pcval = run_pre_commit_hook(stack.repository)
+        if pcval != utils.STGIT_SUCCESS:
+            raise CmdException(
+                'pre-commit hook failed, review the changes using `stg diff`'
+                ', run `stg add` to add them to index and run `stg refresh` again'
+            )
+
```
jpgrayson, line 368: Another behavior that `git commit` has when running `pre-commit` is that it re-reads the index (discarding any cached version of the index) after the `pre-commit` hook is run. This is done to allow for the `pre-commit` hook to modify the index prior to committing. I believe the same need exists in StGit where we would want to allow `pre-commit` to modify files and update the index prior to refresh.

A potential test for this would be to use a `pre-commit` hook that strips trailing whitespace and then does `git add` to update the index with the strip changes, and then returns 0 (success) such that `stg refresh` would capture the stripped whitespace changes.

stgit/commands/common.py
```diff
@@ -461,6 +463,19 @@ def run_commit_msg_hook(repo, cd, editor_is_used=True):
         return cd
 
 
+def run_pre_commit_hook(repo):
+    pre_commit_hook = get_hook(repo, 'pre-commit')
```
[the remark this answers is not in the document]
akshedu: I am not sure if we have to change the `GIT_INDEX_FILE` variable by ourselves. Apparently GIT sets it to a temp index while running hooks and locks the original one as per this stackoverflow answer - https://stackoverflow.com/questions/9437063/how-to-add-a-file-to-the-index-in-a-git-pre-commit-hook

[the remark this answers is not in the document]
akshedu: This is a better explanation of the way git handles `pre-commit` hook in `git commit`.

https://stackoverflow.com/questions/56524769/why-do-pre-commit-hooks-leave-files-half-staged

Look for `More than three copies` in the first answer.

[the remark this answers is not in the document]
akshedu: One more relevant link: https://stackoverflow.com/questions/51587518/whats-the-difference-between-git-commit-file-and-git-commit-only/51587817

stgit/commands/refresh.py
```diff
@@ -354,10 +354,25 @@ def func(parser, options, args):
...
+    if not options.no_verify:
+        pre_commit_success = run_pre_commit_hook(stack.repository)
+        if not pre_commit_success:
+            raise CmdException(
+                'pre-commit hook failed, review the changes using `stg diff`'
+                ', run `stg add` to add them to index and run `stg refresh` again'
+            )
+
+    # Update index and rewrite tree if pre-commit hook updated files in index
+    if not stack.repository.default_index.is_clean(tree):
```
akshedu, line 370: @jpgrayson I am not sure if this is correct. I checked almost all cases and I am getting correct result. But still I have feeling this might be a bug since I am comparing with the default index for checking if hook has changed something.

What do you think?

jpgrayson: I agree that this doesn't feel super robust because of the use of `default_index`. In practice, the default index would only possibly be incorrect when using path limiting, i.e. because that is the only case where a temporary index is used instead of the default index. However, like you, I was unable to craft a test where this code yielded clearly incorrect behavior.

I also spent some time experimenting with `git refresh` and how it's behavior with pre-commit hooks; with and without a pathspec; and with or without changes staged. The semantics of how `git commit` deals with a `pre-commit` script modifying its temp index when using pathspec are non-obvious and, frankly, a bit surprising. It is not clear to me that there is a guiding light from `git refresh` that `stg refresh` should or needs to follow.

In other words, I think this patch as-is is okay. It *might* be more correct to run the pre-commit script in the context of the temp index when path-limiting is in effect (i.e. by setting `GIT_INDEX_FILE`), but again, I do not have a test that shows clearly incorrect behavior when we do not set `GIT_INDEX_FILE` for `pre-commit`.

The one possible change to these two particular lines of code would be to only execute them if we have actually run the pre-commit hook. This would require unrolling `run_pre_commit_hook()` here:

```python
    # Run pre-commit hook, if fails, abort refresh
    if not options.no_verify:
        pre_commit_hook = get_hook(
            stack.repository,
            'pre-commit',
            extra_env={} if options.edit else {'GIT_EDITOR': ':'},
        )
        if pre_commit_hook:
            try:
                pre_commit_hook()
            except RunException:
                raise CmdException(
                    'pre-commit hook failed, review the changes using `stg diff`, '
                    'run `stg add` to add them to index and run `stg refresh` again'
                )
            else:
                # Update index and rewrite tree if hook updated files in index
                if not stack.repository.default_index.is_clean(tree):
                    tree = write_tree(stack, paths, temp_index=path_limiting)
```

And  again, thank you for the thorough work you've done on this feature. It's very much appreciated!

stgit/commands/common.py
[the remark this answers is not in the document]
jpgrayson: In order to set `GIT_INDEX_FILE` for the `pre-commit` script, we would need to call the hook in `write_tree()` and in the context of the temporary `iw` object. This further requires modifications to `get_hook()` to take an `iw` parameter instead of a `repository` parameter.

I've experimented with this idea. The code for `write_tree()` looks something like this:

```python
def write_tree(stack, paths, temp_index, no_verify):
    """Possibly update the index, and then write its tree. If any path
    limiting is in effect, use a temp index.
    @return: The written tree.
    @rtype: L{Tree<stgit.git.Tree>}"""

    def go(index):
        if paths:
            iw = IndexAndWorktree(index, stack.repository.default_worktree)
            iw.update_index(paths)
            pre_commit_hook = get_hook(iw, 'pre-commit')
            if not no_verify and pre_commit_hook:
                try:
                    pre_commit_hook()
                except RunException:
                    raise CmdException(
                        'pre-commit hook failed, review the changes using `stg diff`, '
                        'run `stg add` to add them to index and run `stg refresh` again'
                    )
                index.read_tree(stack.head)
                iw.update_index(paths)
        return index.write_tree()

    if temp_index:
        index = stack.repository.temp_index()
        try:
            index.read_tree(stack.head)
            return go(index)
        finally:
            index.delete()
            stack.repository.default_iw.update_index(paths)
    else:
        return go(stack.repository.default_index)
```

A key thing to take note of is that `IndexAndWorktree.run()` injects `GIT_INDEX_FILE` into the environment. So by supplying the temp `iw` instance to `get_hook()`, it will can run the `pre-commit` script with the temp `iw` instead of its current behavior of using `default_iw`.

Per [my other comment](https://github.com/stacked-git/stgit/pull/63#discussion_r463985880), I do not believe that we must make this change, but since the experiment you describe above failed to use a non-default `GIT_INDEX_FILE`, I thought it might be helpful to illustrate the problem with that experiment.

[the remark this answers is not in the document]
akshedu: Thanks Pete. This was very helpful.

When I first started working on this feature, this was my first thought to break up the `write_tree` function in 2 parts and return index from first part, use that to run hook and write the actually tree only afterwards. Because otherwise the solution is suboptimal. Though I could not come up with a satisfactory solution but your implementation makes sense now.

stgit/commands/refresh.py
akshedu: Thanks Pete. It was a pleasure working on my first open source contribution. If you think there is something relevant for me to implement (features etc) I would be happy to take it up.

Issue #66, opened 2020-08-14
stg import taking minutes on large patches


wberrier:
I have some rather large patches I'm trying to import.

It's been running on a single 28M patch for dozens of minutes.  Running `py-spy top` shows:

```
Total Samples 19600
GIL: 0.00%, Active: 100.00%, Threads: 1

  %Own   %Total  OwnTime  TotalTime  Function (filename:line)
100.00% 100.00%   196.0s    196.0s   __split_descr_diff (stgit/commands/common.py:374)
  0.00%   0.00%   0.010s    0.010s   __split_descr_diff (stgit/commands/common.py:367)
  0.00% 100.00%   0.000s    196.0s   _main (stgit/main.py:193)
  0.00% 100.00%   0.000s    196.0s   parse_patch (stgit/commands/common.py:430)
  0.00% 100.00%   0.000s    196.0s   main (stgit/main.py:221)
  0.00% 100.00%   0.000s    196.0s   <module> (stg:39)
  0.00% 100.00%   0.000s    196.0s   func (stgit/commands/imprt.py:486)
  0.00% 100.00%   0.000s    196.0s   __import_file (stgit/commands/imprt.py:272)
```

This is with v0.23.

Unfortunately, I can't post the patch.  It's a patch that contains changes to 630 files, 700k inserts, some binary files, etc...

It looks like it's getting stuck on parsing the patch, looking for a description...

jpgrayson:
This was a really good find. And thanks again for the repair (#67)!

PR #67, opened 2020-08-15
stg import: optimize string handling for large patches

wberrier:
Referenced from: https://waymoot.org/home/python_string/

This is for issue #66

### DIFF
--- a/stgit/commands/common.py
+++ b/stgit/commands/common.py
@@ -359,19 +359,20 @@ def __end_descr(line):
 def __split_descr_diff(string):
     """Return the description and the diff from the given string
     """
-    descr = diff = b''
+    descr_list = [b'']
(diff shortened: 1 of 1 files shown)


wberrier:
I think I'm also running into this string bottleneck at: `repository.py:DiffTreeProcesses:diff_trees`

jpgrayson:
Thank you for this PR and apologies for the long delay.

This is a really good performance issue to resolve.

I rebased and merged this change in dbc2cf1d86e59b77f0131bb6f8d760d0ac2ae8ab.

But I also made a follow-on change that improves performance further by using a regex to split the patch description and diff.

I have not yet looked deeply into `diff_trees()` and `cat_file()`, but I agree that they seem exposed to the same string concatenation slowness for large diffs/blobs.

PR #73, opened 2020-09-25
stg series -- add numbers and review-tag summaries

jnsnow:
Don't merge this! I didn't test it! I was hoping to get a 2-minute yes/no on:

"This looks useful, please polish this, and add documentation and unit tests", or
"Hmm, I don't think this quite fits with our vision for stgit."

What this does:
(Patches 1-2) Subjective refactoring of series.py to accommodate the two new features easier;
Patch 3: Add '-n' to number patches in the terminal (for cross-referencing against git send-email)
Patch 4: Add an extremely primitive LKML-style review-tag scanner ("Reviewed-by: John Smith <jsmith@example.com>") to show summaries of such tags. For iterative review processes (Hello, QEMU!) where the onus is on the submitter to collect reviews, this is a nice summary report.

As an example (an excerpt of a QEMU series I am working on):
```
> stg series -n -t

+ [01] qapi-expr-py-remove-info       # [SOB] JS [RB] CR,EH
+ [02] qapi-expr-py-check-for-dict    # [SOB] JS [RB] CR,EH
+ [03] qapi-expr-py-constrain         # [SOB] JS [RB] CR,EH
+ [04] qapi-expr-py-add-assertion-for # [SOB] JS [RB] CR,EH
+ [05] qapi-expr-py-move-string-check # [SOB] JS [RB] CR,EH
+ [06] qapi-expr-py-check-type-of     # [SOB] JS [RB] EH
+ [07] qapi-expr-py-add-casts-in-a    # [SOB] JS [RB] CR,EH
+ [08] qapi-expr-py-add-notational    # [SOB] JS [RB] CR,EH
+ [09] qapi-expr-py-rewrite-check_if  # [SOB] JS [RB] CR,EH [TB] CR
```

### DIFF
--- a/stgit/commands/series.py
+++ b/stgit/commands/series.py
@@ -1,4 +1,5 @@
 from collections import namedtuple
+import re
 
 from stgit.argparse import opt, patch_range
 from stgit.commands.common import CmdException, DirectoryHasRepository, parse_patches
@@ -122,6 +123,12 @@ options = [
         action='store_true',
         short='Display the sequence number for each patch',
     ),
+    opt(
+        '-t',
+        '--tags',
+        action='store_true',
+        short='Display some common LKML-style tag indicators',
+    ),
 ]
 
 
@@ -147,6 +154,25 @@ def __get_description(stack, patch):
     cd = stack.patches.get(patch).commit.data
     return cd.message_str.strip().split('\n', 1)[0].rstrip()
 
+_TAG_TABLE = {
+    'RB': 'Reviewed-by',
+    'AB': 'Acked-by',
+    'TB': 'Tested-by',
+    'SOB': 'Signed-off-by',
+    'Rep': 'Reported-by',
+}
+
+def __get_tags(stack, patch):
+    """Extract and return a patch's review-process tags."""
+    tags = {}
+    cd = stack.patches.get(patch).commit.data
+    for line in cd.message_str.split('\n'):
+        for short_tag, tag in _TAG_TABLE.items():
+            match = re.match(tag + ': (.+?) <(.+?)>', line, flags=re.IGNORECASE)
+            if match:
+                initialism = ''.join(word[0] for word in match[1].split(' '))
+                tags.setdefault(short_tag, set()).add(initialism)
+    return tags
 
 def __get_author(stack, patch):
     """Extract and return a patch's short description"""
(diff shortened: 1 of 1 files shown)


jpgrayson:
Thanks for this. I appreciate new ideas for StGit. Some quick impressions:

The `-n` option seems reasonably non-controversial. Although, the `-n/--number` options for `pop` and `push` have a different meaning and take an argument. Maybe not a problem.

Would we want to augment `goto` or any other commands to allow patches to be referenced by numbers?

The tags are less obviously a generally useful feature. I'm not immediately excited about this feature or having to maintain it. I wonder if it might best exist as a separate script? I'd be happy to have it in the `contrib` directory. Alternatively, are there other StGit users (or potential users) that have been wanting this feature? Lots of workflows out in the wild, maybe this scratches an itch a lot of people have (but I don't know about).

I'm okay with refactoring, in general. These refactors seem fine, but I will want to give them more specific review. As the primary active maintainer, I will want them to make sense to me.

So in net, I'm in-favor of what your doing in patches 1-3 and we can move ahead with polishing those, but I'm not sold on the value proposition of 4.

lkraav:
Numbering upgrade is definitely a thing. I constantly struggle with "how many should I push back?" after `stg pop -a` for a pre-patch-queue merge, etc, because we often keep some inactive patches around for ideas and reference.

jnsnow:
Great, thanks for the feedback!

I'll focus on numbering for now. My tag support is a very quick hack, and I understand that these tags are not "part of git" (except the SOB!) in a meaningful way, so supporting it would be very fuzzy. I'll return to thinking about this at a later date, but I am getting good use out of it personally and wanted to share.

I absolutely intentionally broke -short for my WIP, so I need to restore that functionality first. Then I will split my commits down into easier to review pieces. The refactors are not strictly necessary for the support, but if we support more columns, the justification code gets stranger and less accurately named.

I'm open for suggestions on alternatives to `-n`/`--number`, I will trust your judgement on if you would like to prevent recycled flags between subcommands.

It absolutely would be very nice to allow `stg goto 5` or `stg edit 10`, or `stg float 7` -- but these changes suggest a deeper, more fundamental change to the Stack object to give these patches stable and consistent ordering across all subcommands. I can investigate doing this, as I would also find it very useful.

The above might create a new namespace conflict where you cannot name patches 01, 02, 03, etc. I suppose for existing branches, if a patch really truly is named "1" or "05", we can jump to that patch as a preference, but forbid number-named patches going forward. This is probably a reasonable compromise.

I'd also like to add the ability to print `05/47` instead of just `05`, mostly for visual parity with git send-email numbering.

Thanks for taking a quick look, I'll close this PR for now.
--js

Issue #79, opened 2020-11-02
Windows: environment can only contain strings


hborchardt:
Upon install of stgit in a virtualenv in Windows, running `stg show` (or any command) throws the following error:
```
Error: Unhandled exception:
Traceback (most recent call last):
  File "C:\Users\hborchardt\.pyenv\pyenv-win\versions\3.8.3-amd64\lib\site-packages\stgit\main.py", line 188, in _main
    ret = command.func(parser, options, args)
  File "C:\Users\hborchardt\.pyenv\pyenv-win\versions\3.8.3-amd64\lib\site-packages\stgit\commands\series.py", line 194, in func
    stack = directory.repository.get_stack(options.branch)
  File "C:\Users\hborchardt\.pyenv\pyenv-win\versions\3.8.3-amd64\lib\site-packages\stgit\lib\stack.py", line 428, in get_stack
    name = self.current_branch_name
  File "C:\Users\hborchardt\.pyenv\pyenv-win\versions\3.8.3-amd64\lib\site-packages\stgit\lib\git\repository.py", line 223, in current_branch_name  
    return utils.strip_prefix('refs/heads/', self.head_ref)
  File "C:\Users\hborchardt\.pyenv\pyenv-win\versions\3.8.3-amd64\lib\site-packages\stgit\lib\git\repository.py", line 311, in head_ref
    return self.run(['git', 'symbolic-ref', '-q', 'HEAD']).output_one_line()
  File "C:\Users\hborchardt\.pyenv\pyenv-win\versions\3.8.3-amd64\lib\site-packages\stgit\run.py", line 254, in output_one_line
    outlines = self.output_lines(sep)
  File "C:\Users\hborchardt\.pyenv\pyenv-win\versions\3.8.3-amd64\lib\site-packages\stgit\run.py", line 245, in output_lines
    outdata = self._run_io()
  File "C:\Users\hborchardt\.pyenv\pyenv-win\versions\3.8.3-amd64\lib\site-packages\stgit\run.py", line 133, in _run_io
    p = subprocess.Popen(
  File "C:\Users\hborchardt\.pyenv\pyenv-win\versions\3.8.3-amd64\lib\subprocess.py", line 854, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "C:\Users\hborchardt\.pyenv\pyenv-win\versions\3.8.3-amd64\lib\subprocess.py", line 1307, in _execute_child
[...]

PR #80, opened 2020-11-02
Pass environment variables to Popen as dict of strings

hborchardt:
In Python 3, the env argument to Popen has to be a dict of strings, see
https://bugzilla.mozilla.org/show_bug.cgi?id=1585258#c4
This is important when running under Windows, where the unicode strings
are processed directly by the subprocess C code.

However, under posix systems the environment variables are serialized to
bytes using the file system encoding, which may not support unicode
characters.
This means that passing bytes as environment variables and skipping the
encoding is mandatory on posix in order to support unicode environment
variables.

Therefore skip encoding when on Windows.

Fixes: #79

### DIFF
--- a/stgit/run.py
+++ b/stgit/run.py
@@ -88,7 +88,8 @@ class Run:
         return [fsencode_utf8(c) for c in self._cmd]
 
     def _prep_env(self):
-        if self._env:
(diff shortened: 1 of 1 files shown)


jpgrayson:
Thanks for the issue report and this PR! Merged.

PR #81, opened 2020-11-10
stgbashprompt.sh should not be executable

dfaggioli:
It is not necessary for contrib/stgbashprompt.sh to have the executable
permission set.

This was discovered while packaging (for openSUSE) as `rpmlint` reports
this:

  stgit.noarch: W: script-without-shebang /usr/share/stgit/contrib/stgbashprompt.sh
  This text file has executable bits set or is located in a path dedicated for
  executables, but lacks a shebang and cannot thus be executed. If the file is
  meant to be an executable script, add the shebang, otherwise remove the
  executable bits or move the file elsewhere.


jpgrayson:
Thanks for this PR. Merged.

Issue #68, opened 2020-08-19
stg branch --rename forgets to update refs for hidden patches


richardb64:
When renaming a branch using `stg branch --rename`, the refs under `refs/patches` relating to _hidden_ patches are missed.  Any further attempt to use stgit on the affected branch gives a KeyError stacktrace.

In the particular example where I hit this problem, the refs in question (for the _patchname_ itself, and _patchname_.log) were in `.git/packed-refs` rather than `.git/refs/patches`, but that shouldn't be relevant (provided stgit always uses `git update-ref` rather than directly editing files).

### Version details:
Stacked GIT 0.19-1-g2f5bb-dirty
git version 2.26.0
Python version 2.7.5 (default, Sep 26 2019, 13:23:47)

NonLogicalDev:
@richardb64  I tried reproducing this error in tests by:

* Creating patches h0 (hidden) > p0 (popped) > p1 (applied)
* Renaming initial branch frantically =] (for other test cases)
* Renaming branch back to original name
* Renaming branch again and testing `stg id` works for all of the ids of the initially created patches.

That however did not produce any `KeyErrors`, could you double check if this is still an issue for you @richardb64?

Also @jpgrayson, when you get a chance take a look at the slight cleanup of the `stg branch --rename` tests I did along with covering this scenario.

PR #82, opened 2020-11-14
test(branch --rename): Clean up and add tests to cover issue #68

NonLogicalDev:
Attempted to cover reported issue where hidden patches were causing issues with renames. (Issue  #68)


jpgrayson:
Thanks for these test case improvements. Merged.

PR #84, opened 2020-11-20
bug(#83): fix stg pull when no upstream is configured

### LINKED ISSUE #83: `stg pull` on a branch without upstream set is causing an Exception
If a particular branch does not have tracking information spits out traceback.

```
Traceback (most recent call last):
  File "/Users/<USER>/Documents/Developer/stgit/stgit/main.py", line 188, in _main
    ret = command.func(parser, options, args)
  File "/Users/<USER>/Documents/Developer/stgit/stgit/commands/pull.py", line 137, in func
    pull(repository, remote_name)
  File "/Users/<USER>/Documents/Developer/stgit/stgit/commands/pull.py", line 78, in pull
    runner.returns([0, 1]).run()
  File "/Users/<USER>/Documents/Developer/stgit/stgit/run.py", line 265, in run
    self._run_noio()
  File "/Users/<USER>/Documents/Developer/stgit/stgit/run.py", line 161, in _run_noio
    self._prep_cmd(),
  File "/Users/<USER>/Documents/Developer/stgit/stgit/run.py", line 88, in _prep_cmd
    return [fsencode_utf8(c) for c in self._cmd]
  File "/Users/<USER>/Documents/Developer/stgit/stgit/run.py", line 88, in <listcomp>
    return [fsencode_utf8(c) for c in self._cmd]
  File "/Users/<USER>/Documents/Developer/stgit/stgit/compat.py", line 28, in fsencode_utf8
    return os.fsencode(s)
  File "/usr/local/opt/python@3.8/bin/../Frameworks/Python.framework/Versions/3.8/lib/python3.8/os.py", line 806, in fsencode
    filename = fspath(filename)  # Does type-checking of `filename`.
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

NonLogicalDev:
CLOSES  #83

### DIFF
--- a/stgit/commands/pull.py
+++ b/stgit/commands/pull.py
@@ -101,6 +101,7 @@ def func(parser, options, args):
         'stgit.pull-policy'
     )
 
+    remote_name = None
     if policy == 'rebase':
         # parent is local
         if len(args) == 1:
@@ -118,6 +119,12 @@ def func(parser, options, args):
             remote_name = args[0]
         else:
             remote_name = stack.parent_remote
+
+    if policy in ['pull', 'fetch-rebase'] and remote_name is None:
+        parser.error(
+            'There is no tracking information for the current branch.\n'
+            'Please specify which branch you want to merge with.'
+        )
 
     if stack.protected:
         raise CmdException('This branch is protected. Pulls are not permitted')
--- a/t/t1210-pull-check-upstream.sh
+++ b/t/t1210-pull-check-upstream.sh
@@ -1,0 +1,36 @@
+#!/bin/sh
+
+test_description='Check that pulling works when no upstream is configured'
+
+. ./test-lib.sh
+
+# Need a repo to clone
+test_create_repo upstream
+
+test_expect_success \
+    'Setup upstream repo, clone it, and add patches to the clone' \
+    '
+    (cd upstream && stg init) &&
+    stg clone upstream clone
+    (cd clone && git config pull.rebase false)
+    '
+
+test_expect_success \
+    'Test that pull works' \
+    '
+    (cd clone &&
+      git checkout master &&
+      stg pull
+    )
+    '
+
+test_expect_success \
+    'Test that pull without upstream setup produces friendly error' \
+    '
+    (cd clone &&
+      stg branch --create without-upstream &&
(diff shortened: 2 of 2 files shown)


t/t1210-pull.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/sh
+#
+# Copyright (c) 2006 Yann Dirson
```
jpgrayson, line 3: Copy/paste error?

t/t1210-pull.sh
```diff
@@ -0,0 +1,41 @@
...
+      stg pull
+    )
+    '
+
+test_expect_success \
+    'Test that pull without upstream setup produces friendly error' \
+    '
+    (cd clone &&
+      stg branch --create without-upstream &&
+      ( stg pull 2>&1 | ( ! grep -q "Traceback" ) ) || ! stg pull 2>&1
```
jpgrayson, line 37: Instead of testing for the absence of a traceback, I'd prefer we test for the presence of our expected error message.

t/t1210-pull.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/sh
+#
+# Copyright (c) 2006 Yann Dirson
+#
+
+test_description='test
+
+'
```
jpgrayson, line 8: Need better description.

Also, `t1210-pull.sh` is a bit too generic considering there are already four other "pull" test scripts. Please use a more descriptive name for this script.

t/t1210-pull.sh
NonLogicalDev: Removed, artifact of copying another test script.

t/t1210-pull.sh
NonLogicalDev: Agreed, added the description, I missed that field. 

I renamed the test to be more descriptive, though I named it generically just pull because it was testing a generic pull scenario, seems like pull command does not have a generic test to contain common testcases.

t/t1210-pull.sh
NonLogicalDev: Agreed. Fixed

jpgrayson:
Merged. I also pushed a follow-on change to correct a couple things.

Thank you for this PR!

PR #89, opened 2020-12-02
stg series: add option to always show description

### LINKED ISSUE #88: stg series: always show description
There may be some value in allowing users to opt into always showing the
description of patches in `stg series`.

One way to accomplish this is with an `stgit.series.description` configuration
parameter that, when set to `yes` defaults to always showing the description.

To allow disabling the description when this configuration parameter is set,
a `--no-description` flag may also be added.

abhinav:
Add a configuration parameter `stg.series.description` which, when set
to `yes` will always show the description in `stg series`.

With this enabled, users may pass `--no-description` to hide
descriptions in the output.

Resolves #88

### DIFF
--- a/stgit/commands/series.py
+++ b/stgit/commands/series.py
@@ -84,6 +84,12 @@ options = [
         short='Show a short description for each patch',
     ),
     opt(
+        '--no-description',
+        action='store_false',
+        short='Inverse of --description',
+        dest='description',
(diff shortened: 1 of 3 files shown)


jpgrayson:
Thanks for this PR. I think this is a reasonable feature to have. Merged.

PR #90, opened 2021-01-12
Avoid warning message in default pull command

zaneb:
In newer versions of git, a warning message is printed unless one of the
following options is explicitly specified on the command line (or its
equivalent in the config): `--ff` `--no-ff` `--ff-only` `--rebase` `--no-rebase`

To avoid displaying this warning from the internals of git, pass the
`--ff` option in the default pull command.

### DIFF
--- a/stgit/config.py
+++ b/stgit/config.py
@@ -34,7 +34,7 @@ DEFAULTS = [
     ('stgit.pager', ['less']),
     ('stgit.pick.expose-format', ['format:%B%n(imported from commit %H)']),
     ('stgit.pull-policy', ['pull']),
-    ('stgit.pullcmd', ['git pull']),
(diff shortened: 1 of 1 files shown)


jpgrayson:
Adding this to `stgit.pullcmd` will effectively override the user's `pull.ff` and/or `pull.rebase` configuration, if set. I'm not sure StGit should be imposing an override to the user's git config.

Do we believe StGit's pull policy should supersede Git's pull policy? And if so, why is `--ff` the right policy? I.e. why not `--ff-only`?

Given that the git warning gives explicit instructions to the user for how to resolve the warning, my default position is to let this ugly bit of git UI leak through and let StGit users resolve it via the nominal Git mechanisms. But I'm open to being convinced otherwise.

I appreciate this PR. Thanks!

PR #99, opened 2021-03-19
refresh: Support --diff with --edit

### LINKED ISSUE #98: refresh: Support --diff with --edit
I find myself using `--edit` with `stg refresh` often to change the patch description as contents change. Being able to see, and possibly edit the diff, during this would be nice. For vanilla git, I used to have `diff.verbose` set for this reason (minus support for editing). The most obvious way to do this appears to be a `--diff` flag for `refresh` that only has effect if used with `--edit`.

abhinav:
Add a `--diff` flag to `stg refresh` that, when used with `--edit`, is
equivalent to `stg edit --diff`.

Resolves #98

### DIFF
--- a/stgit/commands/refresh.py
+++ b/stgit/commands/refresh.py
@@ -106,6 +106,12 @@ options = (
             short='Invoke an editor for the patch description',
         ),
         opt(
+            '-d',
+            '--diff',
+            action='store_true',
+            short='Edit the patch diff if -e was set',
+        ),
+        opt(
             '-a',
             '--annotate',
             metavar='NOTE',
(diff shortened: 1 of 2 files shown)


jpgrayson:
Want to say thanks for this PR, @abhinav, and sorry I have not yet had time to look at it. I will try to get to it soon.

jpgrayson:
Merged. I appreciate this PR, @abhinav.

That said, I have mixed feelings about blessing `refresh --edit --diff` with the same behavior as `edit --diff`. I agree that _viewing_ the diff inline while editing commit messages is very helpful. BUT, on the other hand, _editing_ diffs along with the commit message is fragile and I consider it an anti-feature of StGit. I have been considering removing the inline diff editing capability. I'm interested in whether anyone relies on this feature in their workflow; and if so, why?

abhinav:
Thanks, @jpgrayson. Yeah, diff editing was a surprising feature. I use this flag to view, not change. Removal of editing makes sense to me if nobody relies on it heavily.

Issue #12, opened 2018-01-13
GPG sign patches on export


terinjokes:
It would be great to my workflow if I could GPG sign patches when exported (eg, with `stg commit`, `stg export`, or `stg mail`). Right now I'm having to `stg commit` the patch, then run `git rebase -S` on it.

lthms:
Yes, please! For work, I need to sign my commits (up to the point where the remote I need to push my commits to won’t accept my commits if they are not correctly signed). This means I cannot really use `stg` in my professional settings, which is a bummer because I really like the workflow it enables D:

Is there a path forward to this? I would happily contribute if being mentored a bit.

jpgrayson:
I am generally in favor of StGit supporting signed commits in some fashion. And I would do my best to mentor if you'd like to take on this feature.

This issue seeks to have patches signed at `stg commit` time, but I tend to think that we would want patches to always be signed when `config.gpgsign=true`, i.e. `new`, `refresh`, and `edit` create signed commits and `commit`, `export`, and `mail` would then just export those already signed commits.

Another question is whether the stack state (a.k.a. log, a.k.a. metadata) commits should be signed?

lthms:
I agree with your point. I guess it would be simpler to systematically follow the git configuration and sign everything that is a commit when git would have signed it.

And I indeed am willing to step in for this feature, since I really enjoy using StGit and would love to be able to use it in a daily basis.

jpgrayson:
There is basically one place in the StGit code where commits are made, in `stgit/lib/git/objects.py`.

https://github.com/stacked-git/stgit/blob/a6b3d497cefaf67d8068e2ae571f730df1a6474e/stgit/lib/git/objects.py#L266-L283

The simplest thing to try would be to use the `-S` option with that `git commit-tree` command. Some challenges:

* How does that work if the user's gpg pinentry program is configured to run in the current tty?
* How do we test this capability?
  - Do we make the test suite conditional on gpg being present (i.e. like we do for quilt)?
  - Are there any tricks to setting up gpg for headless use?
  - How much testing is enough for this feature?
* Do we add the `-S/--gpg-sign` option to various StGit commands? Or do we respect `config.gpgsign`? Or both?
  - I note that the coexistence of `-S` and `--sign` options will be a little awkward. Maybe only support the long `--gpg-sign`?
  - If we support -S/--gpg-sign, do we also support the optional `key-id` argument?
* What about `import`, `pick`, and `fold`? Do these commands sign patches?
[...]

PR #100, opened 2021-04-20
Sign commits when the `commit.gpgsign' git option is set to true

lthms:
This is a first draft related to issue #12. It works as expected, meaning that `stg` effectively signs the commits it creates. I guess this was the easy part, though, as we probably need to refine this to at least provide a `-S`/`--gpg-sign`/`--no-gpg-sign` to the relevant commands.

That being said, @jpgrayson thanks for the first pointers!


jpgrayson:
The next step I'd like to see is some new tests that exercise various StGit commands with `commit.gpgsign = true`. Need to know that StGit doesn't regress when `config.gpgsign = true` in order to be comfortable with adding this change.

We could either weave gpg-enabled tests into existing test scripts, or write one or more new tests scripts to cover gpg signatures with the various subcommands. I think the latter approach would probably be more straightforward.

Another consideration for the tests is whether to require gpg to be installed to run the test suite, or whether to run the gpg-enabled tests conditionally. The prior art for this is for several tests that use `quilt`. See `t/t1802-import-quilt.sh` where all the tests are guarded by the `QUILT` "lazy prereq", which is defined in `t/test-lib.sh`. We could potentially have a similar `GPG` lazy prereq to guard any gpg-enabled tests.

I'll also note that StGit's test suite is based heavily on Git's test suite. For example, StGit's `test-lib.sh` is a lightly modified copy of Git's script of the same name. Looking at how Git's test scripts address gpg and testing signing subcommands may be instructive.

Yet another test consideration is that the GitHub Actions test environment configuration may need to be updated to include `gpg`. See `.github/workflows/ci.yml`. Not sure if the `ubuntu-latest` container we use comes with gpg installed or whether we might need to `apt-get install` it.

lthms:
I’ve pushed a first bunch of tests, so you can have a look and tell me if it is close to what you had in mind (also, you probably need to manually start the CI, as [recently announced by GitHub](https://github.blog/2021-04-22-github-actions-update-helping-maintainers-combat-bad-actors/).

jpgrayson:
Thank you for writing tests, @lthms. I kicked-off CI and it passed.

Seems like the right things happen when we use `-S` in `CommitData.commit()`. I'm generally feeling good about this feature.

So next, I think we need to decide whether to also have command line options. The naming of `-S`/`--gpg-sign`/`--no-gpg-sign` seems okay. The issue, I think, is that a lot of StGit subcommands would need these options added. I wonder if only supporting signatures via the `commit.gpgsign` config option might be a sufficient user experience? And possibly better than many StGit commands being cluttered with these new command line options? What would you think about deferring on adding new command line options and initially only supporting `commit.gpgsign`?

By the way, I'm planning on doing a StGit 1.1 release pretty soon. So there is a decision to be made about whether to try to include this feature in 1.1 or put it into a later release.

lthms:
> What would you think about deferring on adding new command line options and initially only supporting commit.gpgsign?

This makes sense to me. One can locally disable `gpgsign` if they don’t want to use it in a given project.

I think wrapping this up with the current design prior 1.1 is a good idea, then we will be able to move forward and experiments with UX design.

lthms:
(I have updated the test-suite because I have the feeling I was not installing `gnupg2` in the correct location)

jpgrayson:
I think we're agreed to only support `config.gpgsign` for the time being.

I'm happy to merge this PR as-is, but I'd like you to change the PR from "Draft" to "Open" before I do that. Were there any additional changes you were thinking about making before this is merged?

lthms:
Not really, if you are happy with the state of the tests, I would be more than happy to see this merged!

Issue #104, opened 2021-04-29
"stg mail" does not consider the first patch in the series while generating the cover letter diffstats


vlrpl:
`stg mail` doesn't include the first patch in the series to generate diffstats in a cover letter
when `%(diffstat)s` is present.

This is more visible generating a cover letter for a single patch.

Using e.g. the following cover letter template:
```
$ cat cover.txt
From: A U Thor <author@example.com>
Subject: Diffstat Cover Test

A diffstat cover test.

%(diffstat)s
```

The resulting mbox:
`$ stg mail -m --cover=cover.txt $(stg top)`

has an empty `%(diffstat)s` section.

PR #105, opened 2021-04-29
"stg mail": always include the first commit in the cover letter stats

vlrpl:
- Fixes the issue described here: https://github.com/stacked-git/stgit/issues/104 
- Adds a test case to avoid regressions


jpgrayson:
This looks great. Thank you for the issue report and PR, @vlrpl. Merged.

PR #115, opened 2021-06-23
Add --interactive option to rebase

topher200:
`stg rebase --interactive` is a new tool which mirrors `git rebase --interactive`.

Using an editor `--interactive` lets the user reorder patches, squash patch chains, or delete individual patches.

[asciicast](https://asciinema.org/a/BWJZZ65vFtlpGubobNLbixr9G)

Further instructions could be added but this seemed like a good base set. I've been using this locally and it helps execute certain sets of commands much faster! Tests are included.

### DIFF
--- a/stgit/commands/rebase.py
+++ b/stgit/commands/rebase.py
@@ -1,12 +1,22 @@
+import copy
+import itertools
+import re
+import typing
+from enum import Enum
+
+from stgit import utils
 from stgit.argparse import opt
 from stgit.commands.common import (
     CmdException,
     DirectoryGotoTopLevel,
+    delete_patches,
     git_commit,
     post_rebase,
     prepare_rebase,
     rebase,
 )
+from stgit.commands.squash import squash
+from stgit.utils import edit_string
 
 __copyright__ = """
 Copyright (C) 2005, Catalin Marinas <catalin.marinas@gmail.com>
@@ -26,10 +36,10 @@ along with this program; if not, see http://www.gnu.org/licenses/.
 
 help = 'Move the stack base to another point in history'
 kind = 'stack'
-usage = ['[options] [--] <new-base-id>']
+usage = ['[options] [--] [new-base-id]']
 description = """
 Pop all patches from current stack, move the stack base to the given
-<new-base-id> and push the patches back.
+[new-base-id] and push the patches back.
 
 If you experience merge conflicts, resolve the problem and continue
 the rebase by executing the following sequence:
@@ -46,6 +56,12 @@ Or if you want to skip that patch:
 args = ['commit']
 options = [
     opt(
+        '-i',
+        '--interactive',
+        action='store_true',
+        short='Open an interactive editor to manipulate patches',
+    ),
+    opt(
         '-n',
         '--nopush',
         action='store_true',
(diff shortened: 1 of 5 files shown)


topher200:
It seems like the tests for Python 3.5 and 3.6 are broken due to no `@dataclass`. I'll need to implement a replacement.

stgit/commands/rebase.py
```diff
@@ -61,11 +76,156 @@
...
+    DELETE = 1
+    # squash the patch into the previous patch
+    SQUASH = 2
+
+
+@dataclass
+class Instruction:
+    patch_name: str
+    action: Action
+    apply: bool
```
jpgrayson, line 114: Would love to use `dataclass` and type annotations in StGit, but we are currently targeting Python 3.5 as the minimum version, so we cannot use those features yet.

stgit/commands/rebase.py
```diff
@@ -61,11 +77,153 @@
...
+    # open an editor to let the user generate the 'todo' instructions
+    todo = edit_string(
+        '\n'.join(instructions),
+        '.stgit-rebase-interactive.txt',
+    )
+
+    # parse the instructions we've been given
+    seen_apply_line = False
+    instructions = []
+    for line in todo.split('\n'):
```
jpgrayson, line 146: Nit: `todo.splitlines()`

stgit/commands/rebase.py
```diff
@@ -61,11 +77,153 @@
...
+        line = line.strip()
+        if line == INTERACTIVE_APPLY_LINE:
+            seen_apply_line = True
+
+        # ignore comment lines
+        if not line or line.startswith('#'):
+            continue
```
jpgrayson, line 153: A keyboarding mistake I sometimes make is to type `J` in vim, which concatenates the subsequent line with the current line. I'm thinking about a case where the user's instructions end up like this:

```
keep p0
# --- APPLY LINE ---keep p1
keep p2
```
So I'm wondering if it's worth the trouble of searching for the apply line more robustly such that we can warn or error-out if the user puts junk after (or before) the apply line string?

I'm not sure anything must be done about this; just want you to consider it.

stgit/commands/rebase.py
```diff
@@ -74,14 +232,19 @@ def func(parser, options, args):
...
-
     rebase(stack, iw, target)
-    if not options.nopush:
-        return post_rebase(stack, applied, 'rebase', check_merged=options.merged)
+
+    if options.interactive:
+        return __do_rebase_interactive(repository, applied, check_merged=options.merged)
+    else:
+        if not options.nopush:
+            return post_rebase(stack, applied, 'rebase', check_merged=options.merged)
```
jpgrayson, line 250: Nit: reduce nesting with `elif`.

jpgrayson (approved):
Thanks for implementing this feature. I'm suspect this will be valuable to many StGit users.

I've added a few review comments for you to consider, but I think this code is good enough as-is. Thank you also for writing tests!

stgit/commands/rebase.py
topher200: done

stgit/commands/rebase.py
topher200: Fixed! Added a test for it.

```
        # record when we find the APPLY_LINE so we know which patches to apply
        if INTERACTIVE_APPLY_LINE in line:
            if INTERACTIVE_APPLY_LINE == line:
                seen_apply_line = True
            else:
                raise CmdException("Bad APPLY_LINE: '%s'" % line)
``

stgit/commands/rebase.py
topher200: :+1:

topher200:
> Thank you also for writing tests!

The test suite is what drew me to write this as a first-class citizen in the first place! (as opposed to a separate script in `contrib`). Once I got the hang of the harnesses I found it very easy to use a TDD style for adding new features to rebase.

I'd love any feedback you have on how to clean up the tests. I went for verbosity over DRY whenever I was confused, which is great because it led to landing the feature quickly :joy:. Happy to clean them up however you see fit.

One (potential) downside of the way the `--interactive` tests are currently architected is that we can only set `fake-editor` once. That means that if two editor windows are opened by a single command (easy to do: write a `stg rebase --interactive` which includes a `squash` instruction, example: [1]), we don't have any control over that second editor (in this case, the `squash`). The `git` repo gets around this restriction by having _two_ fake-editors and somehow using the second one for the squash, but I didn't go there. Instead I just wrote my tests such that the commit message of the squash patch isn't relevant to the test. I don't think it's a big deal, just something I wanted to mention.

I have a few test steps like this: `    stg delete $(stg series --all --noprefix --no-description) &&`. I needed to delete all patches to reset my stack (mostly due to the editor issue above :joy:). I'm surprised there wasn't a lib function to do this already -- feel free suggest an alternative way of refreshing the stack (or creating a new stack) in the middle of a suite of tests.

[1]: https://github.com/stacked-git/stgit/pull/115/files#diff-ec771ac340142cbdf8090d75cd33e3c825b4be6d2b43d1055e2482e5863908bcR131-R139

PR #116, opened 2021-06-24
Rearchitect squash edit message to include all commits.

### LINKED ISSUE #71: Feature request: Add option to stg squash to preserve text below the `---` line
When using `stg squash` and editing the patch description, I'd like `stg squash` to preserve the text below the `---` line, because I keep a list of changes between the patch's published revisions there.
Currently I need to replace `---` with something else before I finish editing in `stg squash` and later change it back with `stg edit`, but it would be more convenient if I could pass an option to `stg squash` that would cause it to perserve the text.

topher200:
Resolves #71.

This change makes stgit's squash commit edit message mirror git's.

[asciicast](https://asciinema.org/a/uke1YRL44VOAOqHkUmcqxItS3)

New edit message example:
```
# This is a combination of 3 patches.
# This is the commit message for foo (patch #1):
foo

# This is the commit message for bar (patch #2):
bar

# This is the commit message for baz (patch #3):
baz

# Please enter the commit message for your patch. Lines starting
# with '#' will be ignored.
```

### DIFF
--- a/stgit/commands/squash.py
+++ b/stgit/commands/squash.py
@@ -61,24 +61,10 @@ class SaveTemplateDone(Exception):
     pass
 
 
-def _append_comment(message, comment):
-    return '\n'.join(
-        [
-            message,
-            '',
-            '---',
-            'Everything following the line with "---" will be ignored',
-            '',
-            comment,
-        ]
-    )
-
-
-def _strip_comment(message):
-    try:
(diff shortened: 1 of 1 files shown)


jpgrayson:
Wow, thanks for this PR. Let's make this the new approach for squash!

PR #117, opened 2021-06-24
Ensure a newline after each commit in 'stg squash'.

topher200:
I observed an issue locally where if a commit message does not include a
trailing newline, the suggested commit message is more "scrunched" than
we'd like it to be -- there's no newline between the commit message and the
instructions.

squash message

This scenario is easy to get into if you perform a squash followed by
another squash onto the same patch (by default the trailing newline is 
stripped).

The fix here is to strip all newlines from the end of each commit
message and then add two newlines to the end (one to get to the next
line, another to ensure the blank line between messages).

### DIFF
--- a/stgit/commands/squash.py
+++ b/stgit/commands/squash.py
@@ -1,3 +1,4 @@
+# TODO (not this file): add 'verbose=1' support to 'stg new', similar to 'git commit'
 from stgit import argparse, utils
 from stgit.argparse import opt, patch_range
 from stgit.commands.common import (
@@ -86,7 +87,7 @@ def _squash_patches(trans, patches, msg, save_template, no_verify=False):
                 pn,
                 num,
             )
-            msg += "\n%s\n" % trans.patches[pn].data.message_str
+            msg += "\n%s\n\n" % trans.patches[pn].data.message_str.strip()
         msg += (
             "# Please enter the commit message for your patch. Lines starting\n"
             "# with '#' will be ignored."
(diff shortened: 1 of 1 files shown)


stgit/commands/squash.py
```diff
@@ -1,3 +1,4 @@
+# TODO (not this file): add 'verbose=1' support to 'stg new', similar to 'git commit'
```
jpgrayson, line 1: This looks out of place.

stgit/commands/squash.py
```diff
@@ -86,7 +87,7 @@ def _squash_patches(trans, patches, msg, save_template, no_verify=False):
                 pn,
                 num,
             )
-            msg += "\n%s\n" % trans.patches[pn].data.message_str
+            msg += "\n%s\n\n" % trans.patches[pn].data.message_str.strip()
```
jpgrayson, line 90: Have you considered whether `rstrip()` might be more appropriate than `strip()`? I.e. Should we respect any leading whitespace in a user's commit message?

A concrete example:

```
    Indented message header line

Rest of message.
```

`strip()` and `rstrip()` are going to give different outcomes for the above case. My general sentiment is that it's safe and commendable to normalize the trailing newlines of commit messages, but messing with other user-provided whitespace risks breaking someone's workflow.

jpgrayson (changes requested):

stgit/commands/squash.py
topher200: good catch! i've moved my TODO list to an external doc :p

stgit/commands/squash.py
topher200: Totally agree, will fix

PR #118, opened 2021-06-24
Reorder squash message to ease readibility.

topher200:
After using this in practice, I think that having the patch number
followed by the patch name is more readable. With this change it's
easier to scan the commit message to determine where each change
ends and the next one begins.

Before:
before

After:
image

I recognize that this PR is more personal preference; it's not the end
of the world if we leave it as-is.


jpgrayson:
I like this updated form better too. Glad you're optimizing this.

Another possible modification to consider: show the squashed patch name somewhere. Possibly on the first line. E.g.:

```# squashed-patch-name is a combination of N patches```

or

```# Squashing N patches as squashed-patch-name```

The latter would be better if reusing the name of one of the patches being squashed (which I often do).

PR #120, opened 2021-06-24
Include 'name' in squash's edit message.

### LINKED ISSUE #119: Include name of squashed patch in squash edit message.
A possible modification to consider in the `stg squash` edit message: show the squashed patch name somewhere. Possibly on the first line. E.g.:

```# squashed-patch-name is a combination of N patches```

or

```# Squashing N patches as squashed-patch-name```

The latter would be better if reusing the name of one of the patches being squashed (which I often do).

_Originally posted by @jpgrayson in https://github.com/stacked-git/stgit/issues/118#issuecomment-867903919_

topher200:
Include 'name' in squash's edit message.

If --name isn't included we don't try to guess what the name may be
(since the name is dependent on the commit description the user gives to
the squashed commit.

Tested locally that this works correctly with and without '--name'.

[asciicast](https://asciinema.org/a/qZb3YV9KwN95zctV1QnLnuQQZ)

Resolves #119

### DIFF
--- a/stgit/commands/squash.py
+++ b/stgit/commands/squash.py
@@ -67,7 +67,7 @@ def _strip_comments(message):
             yield line
 
 
-def _squash_patches(trans, patches, msg, save_template, no_verify=False):
+def _squash_patches(trans, patches, name, msg, save_template, no_verify=False):
     cd = trans.patches[patches[0]].data
     for pn in patches[1:]:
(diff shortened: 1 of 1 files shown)


jpgrayson:
Thanks!

Also, I appreciate the asciinema demos you're doing.

PR #125, opened 2021-07-05
Add --autostash to 'stg rebase'

topher200:
This mirrors a feature of 'git rebase' where, instead of erroring on a dirty
worktree, `--autostash` will stash the changes and apply them after the rebase
has completed.

I'm not sure if it's more 'stgit-ish' to create a patch or a stash. I kept the
git behavior (a stash) but I could see arguments either way.

[asciicast](https://asciinema.org/a/423971)

### DIFF
--- a/stgit/commands/rebase.py
+++ b/stgit/commands/rebase.py
@@ -16,6 +16,8 @@ from stgit.commands.common import (
     rebase,
 )
 from stgit.commands.squash import squash
+from stgit.config import config
+from stgit.run import RunException
 from stgit.utils import edit_string
 
 __copyright__ = """
@@ -73,6 +75,18 @@ options = [
         action='store_true',
         short='Check for patches merged upstream',
     ),
+    opt(
+        '--autostash',
+        action='store_true',
+        short='Stash changes before the rebase and apply them after',
+        default=config.getbool('stgit.autostash'),
+        long='''
+    Automatically create a temporary stash before the operation begins, and
+    apply it after the operation ends. This means that you can run rebase on a
+    dirty work-tree. However, use with care: the final stash application after a
+    successful rebase might result in non-trivial conflicts.
+    ''',
+    ),
 ]
 
 directory = DirectoryGotoTopLevel()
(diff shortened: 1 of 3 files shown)


stgit/commands/rebase.py
```diff
@@ -242,13 +256,28 @@ def func(parser, options, args):
     else:
         target = stack.base
 
+    if options.autostash and not iw.worktree_clean():
+        Run('git', 'stash').run()
```
jpgrayson, line 260: We should prefer using `repository.run()` instead of `Run()`. The former sets the `GIT_DIR` environment variable to give Git commands a normalized environment.

Also, I'd also prefer if we used the more explicit `git stash push` here.

jpgrayson (changes requested):
I think this is a great feature.

Regarding using stash versus creating a temporary patch, the difference in those approaches will only be material when the stashed changes do not apply cleanly after the rebase. Using stash is a more straightforward approach to leaving changes in the working tree, but will require the stgit user to understand what to do with a failed stash pop. I.e. they will have to:

1. Resolve conflicts
2. Call `git stash drop` to eliminate the stash

Would it make sense to wrap our invocation of `git stash pop` such that if it fails we can provide the above instruction to the user?

If we went with the alternative approach of creating a temporary patch and the post-rebase application of the temp patch failed due to conflicts, the user would be left with conflicts in the work tree (just like with stash) along with a `rebase-temp` patch at the top of the stack. In this scenario, the user would have to:

1. Resolve conflicts
2. Call `stg refresh`
3. Decide what to do with temporary `rebase-temp` patch:
    a. Call `stg delete --spill rebase-temp` to leave changes in the working tree
    b. Call `stg rename` to rename `rebase-temp` to something else.

So, my interpretation is that stash is going to provide a better experience, but we should provide some instruction for the user in the case of conflicts.

topher200:
For reference, here's the current text output when the `git stash pop` fails:

```
$ sri --autostash
Saved working directory and index state WIP on master: 4f19d50 new
Popping all applied patches ...
  Popped new
done
Rebasing to "0e0e902f868923f8cc1da7f277593ec4a74b9468" ... done
Invoking the editor: "vim .stgit-rebase-interactive.txt" ... done
Auto-merging one.txt
CONFLICT (content): Merge conflict in one.txt
Recorded preimage for 'one.txt'
The stash entry is kept in case you need it again.
stg rebase: git failed with code 1
```

Not very helpful!

---

Updated text:

```
Rebasing to "0e0e902f868923f8cc1da7f277593ec4a74b9468" ... done
Invoking the editor: "vim .stgit-rebase-interactive.txt" ... done
Auto-merging one.txt
CONFLICT (content): Merge conflict in one.txt
Recorded preimage for 'one.txt'
The stash entry is kept in case you need it again.
stg rebase: Merge conflict when popping stack after rebase. Resolve the
            conflict then delete stash with `git stash drop`.
```

Happy to improve it further if anyone has suggestions!

jpgrayson:
Thank you!

PR #128, opened 2021-07-05
Resolve 'flake8' errors.

topher200:
Seeing these errors when running flake8 locally:

```
make lint-flake8
./stgit/utils.py:78:47: B006 Do not use mutable data structures for argument defaults.  They are created during function definition time. All calls to the function reuse this one instance of that data structure, persisting changes between them.
./stgit/commands/mail.py:467:52: B006 Do not use mutable data structures for argument defaults.  They are created during function definition time. All calls to the function reuse this one instance of that data structure, persisting changes between them.
./stgit/commands/uncommit.py:123:13: B007 Loop control variable 'i' not used within the loop body. If this is intended, start the name with an underscore.
./stgit/commands/__init__.py:90:14: B007 Loop control variable 'mod' not used within the loop body. If this is intended, start the name with an underscore.
```

It's not clear to me why the CI isn't failing with the same errors.

### DIFF
--- a/stgit/commands/__init__.py
+++ b/stgit/commands/__init__.py
@@ -87,7 +87,7 @@ def py_commands(commands, f):
 
 def _command_list(commands):
     kinds = {}
-    for cmd, mod, kind, help in commands:
+    for cmd, _, kind, help in commands:
         kinds.setdefault(kind, {})[cmd] = help
     for kind in _kind_order:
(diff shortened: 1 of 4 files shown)


jpgrayson:
Looks like your local environment has `flake8-bugbear` installed. Perhaps that should be explicitly part of StGit's lint suite.

PR #134, opened 2021-07-07
Add coveralls.io link to README.

topher200:


jpgrayson:
I'm not 100% sold on advertising the code coverage in the readme.

That said, there are currently no links to the coveralls.io page anywhere to be found.

What if instead we added a statement about code coverage and the link to coveralls.io to `CONTRIBUTING.md`?

topher200:
I've been surprised how difficult it is to get coverage data for individual commits. I mean, I can run `make coverage` but then knowing how that compares against `master` is a tedious manual process. That sucks! I looked into how to add coveralls to each PR but I don't see a way to do it without leaking the API secret (which you clearly ran into as well, seeing as it's currently only running on `master`).[1]

I don't have a great solution either! I think it's great that the tool has >90% coverage and coveralls has been great to know what branches need tests before I can refactor them; I never would have been able to land https://github.com/stacked-git/stgit/pull/133 and https://github.com/stacked-git/stgit/pull/137 without first writing https://github.com/stacked-git/stgit/pull/132, and the only reason I knew to write https://github.com/stacked-git/stgit/pull/132 was because of the coveralls. So that was very cool! I've never used a project that had a `make coverage` command and I'm loving it's presence.

So my thought is to make coveralls.io as visible as possible, even if we can't have it on each PR. I don't quite understand your aversion to having it on the README but I'm not here to change your mind. Happy to switch this PR over to `CONTRIBUTING.md` instead.

[1] Ok -- I thought of a way to do it. Require a maintainer to OK a PR, which can then trigger the CI to run any Actions which contain secrets. That way there's a check against someone dumping the secret in their PR.

jpgrayson:
> I don't quite understand your aversion to having it on the README

I expect the README's audience to primarily be people who may be interested in _using_ StGit. The coveralls.io badge with its coverage number _might_ be a confidence builder for the StGit-curious, but ultimately it's secondary to the description of what StGit is, how to install it, and where to learn more about it. But the badge ends up being the very first thing the reader will see, as well as being the most colorful thing. As happy/proud I am of StGit's coverage, it's just not the main message I want to convey via the README.

Also, this would be the only badge. My opinion might tip the other way if we had several badges that conveyed the overall health of the project.

topher200:
Closed in favor of https://github.com/stacked-git/stgit/pull/139

PR #137, opened 2021-07-09
Edit gains the power to rename the patch.

topher200:
'stg edit' now has a new field at the top of the edit window:

    'Patch: <patch_name>'

Editing this value will result in the patch being renamed. If you 
instead 'blank' this field, it will re-generate the patch name using 
the commit message.

[asciicast](https://asciinema.org/a/424730)


jpgrayson:
This is a feature I will use. Thanks!

That said, I think a problem will be discoverability. Not sure what, if anything, should be done about that.

PR #122, opened 2021-06-27
Add --verbose option to 'stg new'

topher200:
Git has a great feature where you can get a 'git diff' in the editor window when
creating a commit. This can help give context while you write your commit
message.

This PR adds this functionality to 'stg new'. Just like Git, there's two ways to
trigger the new feature:
1. `stg new --verbose`
2. Setting the config setting `commit.verbose=1` (borrows Git's config var)

[asciicast](https://asciinema.org/a/422507)

PR #141, opened 2021-07-14
new: commit.verbose is a boolean

abhinav:
In #122, `new` got a `--verbose` flag that defaults to true if the
`commit.verbose` configuration in git is set. To determine whether the
configuration was set, the new feature used `config.getint`.

However, per the git-config documentation, `commit.verbose` can be a
boolean:

       commit.verbose
           A boolean or int to specify the level of verbose with git commit. See git-commit(1).

In fact, if `commit.verbose` is set to `true` right now, this will fail with,

    Value for "commit.verbose" is not an integer: "true"

Given that `config.getbool` will handle both, integers and booleans, use
that to determine if the configuration is set instead. Since
`config.getbool` will return `None` if unset, default to False rather than
None.

### DIFF
--- a/stgit/commands/new.py
+++ b/stgit/commands/new.py
@@ -88,8 +88,7 @@ def func(parser, options, args):
     if options.verbose:
         verbose = options.verbose
     else:
-        verbose_int = config.getint('commit.verbose')
-        verbose = verbose_int > 0 if verbose_int else False
+        verbose = config.getbool('commit.verbose') or False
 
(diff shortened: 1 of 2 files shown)


jpgrayson:
This change makes sense to me.

@topher200, your thoughts?

topher200:
Shipit!

topher200 (approved):

jpgrayson:
Thanks @abhinav!

Issue #138, opened 2021-07-09
Make all the great features of `stg edit` more discoverable


topher200:
Migrating discussion from https://github.com/stacked-git/stgit/pull/137. cc @jpgrayson 

`stg edit --diff` can do a lot of things:
1. rename the patch
2. change the author
3. change the date
4. change the commit message
5. change the diff(!!!)

Basically, everything on the page is editable. How can we make this more discoverable?

---

The "dumb" idea would be to add instructions, similar to `stg rebase --interactive` or `git commit`.

### `git commit`
![image](https://user-images.githubusercontent.com/206988/125141690-be4f5400-e0e3-11eb-8f85-b30ac3e5b8ee.png)

### `stg rebase --interactive`
![image](https://user-images.githubusercontent.com/206988/125141712-cb6c4300-e0e3-11eb-86bd-e3811b9fef2c.png)

---

We could drop some text in the middle of the edit message. 

### `stg edit --diff` currently:
![image](https://user-images.githubusercontent.com/206988/125141796-040c1c80-e0e4-11eb-8553-35923b461226.png)

Proposal: how about something like this?

![image](https://user-images.githubusercontent.com/206988/125142020-a75d3180-e0e4-11eb-918d-2b6a080e2dcf.png)

```Patch: baz-new
From: Topher Green <topher@memfault.com>
Date: 2021-07-05 18:23:42 -0400

baz

---
# Everything in `edit` is editable! Modify the patch name, author, date,
# commit message and (if `--diff` was given) the diff below.

 one.txt |    2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/one.txt b/one.txt
index de9f81b..c70b308 100644
--- a/one.txt
+++ b/one.txt
@@ -1,4 +1,4 @@
-First line
+First line new
 Second line
 Third line
 Fourth line
```

I have no ego about any of this text -- happy to iterate if anyone has suggestions.

AlansCodeLog:
Didn't realize it was so powerful. Awesome!

Regarding the message:
- Does it have to be in the middle?
- Also this part reads weird/confusing to me "Everything in `edit` ...". My mind immediately went to what property edit? Then I realized it was talking about the command.

I think something like this at the top is clearer:

```
# Everything here is editable! You can modify the patch name, author,
# date, commit message, and the diff (if `--diff` was given).
# Lines starting with # will be ignored.
```
For the last line, I copied the start from git, but does emptying it abort the edit like git too or ???

[...]

PR #143, opened 2021-07-18
'stg edit' aborts on empty patch description; add instructions.

topher200:
This resolves https://github.com/stacked-git/stgit/issues/138 by...

1. Adding instruction hints on the *power* of 'stg edit'.
2. Stripping these new hint lines from the description.
3. Aborting 'stg edit' if all description lines are cleared.

From conversation on https://github.com/stacked-git/stgit/issues/138 it became
clear that aborting the 'edit' on an empty editor is correct.  This behavior
mirrors 'git commit' and 'stg squash'.

Previously, if a user cleared the 'stg edit' editor window, the patch
description would be cleared.  This is unintuative! Users likely expect that
this would abort the edit instead.

[asciicast](https://asciinema.org/a/426044)

### DIFF
--- a/stgit/commands/common.py
+++ b/stgit/commands/common.py
@@ -369,10 +369,14 @@ def __split_descr_diff(string):
     return desc, diff
 
 
-def __parse_description(descr):
+def __parse_description(descr, fail_on_empty_description):
     """Parse the patch description for patch and author information."""
     subject = ''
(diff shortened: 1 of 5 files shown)


jpgrayson:
Thanks for this PR. Nice refinements!

PR #146, opened 2021-07-30
Add test case for `stg new` without `--verbose`.

topher200:
Add test cases to show that `stg new` does not include a diff with `--verbose`.

I expanded the existing test cases to check that there's actually a diff included in the commit message editor window for `--verbose`; I confirm there is no diff when there is no `--verbose`.

### DIFF
--- a/t/t1003-new.sh
+++ b/t/t1003-new.sh
@@ -87,27 +87,45 @@ test_expect_success \
     stg show | grep "Patch Description Template"
 '
 
+test_expect_success 'Setup fake editor' '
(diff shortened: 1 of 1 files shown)


topher200:
This PR originally suggested that there was an expected test failure but then I realized I was incorrectly checking the result of the 'grep' -- I no longer think there's any expected errors.

jpgrayson:
Thank you for this fix, @topher200. As always, much appreciated. Sorry for the delay in merging!

PR #147, opened 2021-07-30
Fix off-by-one-error for 'stg rebase -i delete'

topher200:
Previously, we were using itertools.count(). That's great! But we also were
performing an `index -= 1` in the loop which was getting ignored.

The new algorithm explicitly deletes the patch from the instruction set and
correctly counts index on deletes.

Added a test case which proves the off-by-one error is fixed and that deletes
work correctly.

### DIFF
--- a/stgit/commands/rebase.py
+++ b/stgit/commands/rebase.py
@@ -238,13 +238,15 @@ def __do_rebase_interactive(repository, previously_applied_patches, check_merged
             Instruction(patch_name, instruction_type, not seen_apply_line)
         )
 
-    for index in itertools.count():  # pragma: no branch
+    index = 0
+    while True:  # pragma: no branch
         if index >= len(instructions):
             break  # reached the end of the instruction list
         patch_name = instructions[index].patch_name
         if instructions[index].action == Action.DELETE:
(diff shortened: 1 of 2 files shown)


stgit/commands/rebase.py
```diff
@@ -267,10 +269,16 @@ def __do_rebase_interactive(repository, previously_applied_patches, check_merged
...
 
         instructions = __perform_squashes(instructions, index, stack)
+        index += 1
+
+    post_rebase(
+        stack,
+        [instruction.patch_name for instruction in instructions if instruction.apply],
+        'rebase',
+        check_merged,
+    )
```
topher200, line 281: I was trying to think of why I didn't do this in the first place (`apply` all the patches once at the end instead of doing them individually during the loop). 

Then I remembered that until https://github.com/stacked-git/stgit/pull/127, there was no way to `apply` a patch which had been `squashed` into as part of the same instruction set... the squash patch has the potential to be renamed as part of the squash! However with the addition of https://github.com/stacked-git/stgit/pull/127 we can be sure that we still have a 'valid' instruction set at this point -- all instructions which might modify a patch name (ie `squash` and `edit`) now update the `instructions` set to use the new patch name.

PR #148, opened 2021-08-04
completion: Add --no-description for stg series

abhinav:
If the `stg.series.description` config is set to `true`, then the output of
`stg series` always includes the first line of the commit messages.

For example,

    $ stg series
    + foo # Add foo
    + bar # Rename bar
    > baz # Remove baz

When the completion scripts call `stg series`, they use the full string,
including the description:

    $ stg refresh -p <tab>
    foo\ \#\ Add\ foo

To solve this, pass the `--no-description` flag to `stg series` in all
completion scripts when retrieving a list of patches.

### DIFF
--- a/completion/stgit.zsh
+++ b/completion/stgit.zsh
@@ -714,7 +714,7 @@ __stg_patches_all() {
     declare -a patches
     local expl
     patches=(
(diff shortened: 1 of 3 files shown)


jpgrayson:
Thank you for this PR, @abhinav ! And sorry for the delay getting it merged.

PR #149, opened 2021-08-07
'stg rebase --interactive' learns 'fixup' instruction

topher200:
'fixup' can be used to 'squash' a patch into the previous patch BUT keeping
only the base patch's commit message. It's a faster 'squash' for the times you
know you have an awesome commit message already.

[asciicast](https://asciinema.org/a/s5IfTxhE0Yl0SrHxDN1xkZI2s)


jpgrayson:
This looks like another nice improvement to interactive rebase. Thanks, @topher200.

PR #165, opened 2021-11-23
WIP/DO_NOT_MERGE: Extend note copy procedure to respect notes.rewriteRef

### LINKED ISSUE #108: Extension STG to provide configuration should respect `notes.rewriteRef`
## Problem:

STG is using `git notes copy <old> <new>` command to copy over notes upon modifying stack.
This means that users only option is to set `core.notesRef` (if the noted they want to preserve are anything other than `refs/notes/commits`)
But even this workaround allows to carry over only one ref, all others will be lost.

## Background:

This is a fairly niche problem, but some tools use notes to attach metadata to commits, and currently the best course of action is to set `core.notesRef` to the notes ref you care about the most. And abandon tools that rely on other refs.

Proposed Solution:

STG should respect `notes.rewriteRef` configuration. 
It is a multi value configuration that allows user to instruct git to carry over notes during rebase and amend operations.
It also supports glob expressions.

```
$ git config --get-all notes.rewriteRef
refs/notes/commits
refs/notes/farc
refs/notes/stg-*
```

Git has an option `--for-rewrite` on `git notes copy`, which is, bizarrely, undocumented.

https://github.com/git/git/blob/7e391989789db82983665667013a46eabc6fc570/builtin/notes.c#L498

You can also find examples in tests:
https://github.com/git/git/blob/7e391989789db82983665667013a46eabc6fc570/t/t3301-notes.sh#L1117

This option comes with a small catch that it forces `git notes copy` into reading pairs of references from stdin.

So rather than:
```
git notes copy <old> <new>
```

The correct way to invoke this would be:
```
echo "<old> <new>" | git notes copy --for-rewrite=stg
```

This has a side effect of letting the user configure merge strategy via `notes.rewrite.<stg> = true|false`.

## Code that needs to be tweaked:

https://github.com/stacked-git/stgit/blob/68f949852ab7f0e8bb52c6a6fc2ece2a74ded824/stgit/lib/git/repository.py#L472-L475

NonLogicalDev:
CLOSES #108

### DIFF
--- a/stgit/lib/git/repository.py
+++ b/stgit/lib/git/repository.py
@@ -483,5 +483,9 @@ class Repository:
 
     def copy_notes(self, old_sha1, new_sha1):
         """Copy Git notes from the old object to the new one."""
-        p = self.run(['git', 'notes', 'copy', old_sha1, new_sha1])
+        p = self.run(['git', 'notes', 'copy', '--for-rewrite=stg'])
+        p.env({
+            "GIT_NOTES_REWRITE_REF": "refs/notes/*"
+        })
+        p.raw_input("{} {}".format(old_sha1, new_sha1))
         p.discard_exitcode().discard_stderr().discard_output()
--- a/t/t2700-refresh.sh
+++ b/t/t2700-refresh.sh
@@ -4,6 +4,73 @@ test_description='Run "stg refresh"'
 
 . ./test-lib.sh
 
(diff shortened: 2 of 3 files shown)


jpgrayson:
I'm closing PRs for the old Python implementation. This will need to be resurrected for the Rust implementation.

PR #169, opened 2021-12-01
Change os.system() call to avoid invoking 'python', which isn't always available

vsajip:
Some systems (e.g. recent Ubuntu, Linux Mint) don't have a `python` command, but only `python2` or `python3`. This change invokes `os.system()` with the currently running interpreter, which avoids an error that occurs and is displayed when `python` is not available.


jpgrayson:
Good fix. Thanks for this, @vsajip.

PR #171, opened 2021-12-13
stack: strip trailing dots from short_name

MaxKellermann:
Fixes assertion failure which can occur because the short_name is a
truncated version of the already-sanitized long_name, but the
truncation may leave a trailing dot.


jpgrayson:
Good find. Thanks for this PR @MaxKellermann.

Issue #187, opened 2022-05-22
diff-opts can no longer be used multiple times


svenwest:
stg 2.0.0-alpha.0

I use the stgit emacs mode, and it fails to show the diff of a patch with the following error:
```
error: The argument '--diff-opts <OPTIONS>' was provided more than once, but cannot be used multiple times
```

It seems rust-stg expects a space-separated string, while legacy-stg expects multiple occurences of the `-O` argument.

svenwest:
I tried to create this under the label 'rust' but I can't edit the labels, and there's no label 'rust'.

PR #189, opened 2022-05-22
Multiple diff-opts

svenwest:
Solves Issue #187
Not super-happy about adding the binary-argument to the diff_tree_patch function but if we can have an argument for `--color` I suppose we can have one for `--binary` too and it makes the handling of `--diff-opts` a lot simpler.

### DIFF
--- a/src/argset.rs
+++ b/src/argset.rs
@@ -29,6 +29,7 @@ lazy_static! {
         .help("Extra options to pass to \"git diff\"")
         .takes_value(true)
         .allow_hyphen_values(true)
+        .multiple_occurrences(true)
         .value_name("OPTIONS")
         .value_hint(ValueHint::Other);
 }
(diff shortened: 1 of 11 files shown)


jpgrayson:
Thanks for this PR @svenwest. First Rust PR award goes to you! :trophy: 

I'm working on some follow-on patches that resolve the clippy issue and allow both multiple `--diff-opts` (as this patch enables) and a single `--diff-opts` containing multiple space-separated options. The Python implementation works this way, so supporting both in the Rust implementation should maximize compatibility.

Also currently missing from the Rust implementation is comprehension of the `stgit.diff-opts` configuration. I've got a patch for that too.

Anyway, I really appreciate that you're trying out the Rust implementation of StGit and doubly appreciative of the PR. Thanks!

svenwest:
Thanks! :partying_face: 
Great work on the other two patches, looks much better than with just my quick-fix.

I've been using StGit daily in my work for the past ~7 years so I don't mind contributing when I have the chance.

PR #199, opened 2022-08-12
Use first() instead of get(0) as required by clippy 0.1.63

proski:
New rust, new clippy, new things to fix.
This needs to be fixed for automated checks to pass.
Tested by `make lint` locally with up-to-date Rust (rustup update stable).


jpgrayson:
Merged in [7c4a742](https://github.com/stacked-git/stgit/commit/7c4a742d3055430a1943dea0e5957320af8754e2)

Thank you for this PR, @proski.

PR #201, opened 2022-08-12
Use "patch" extension when editing a patch with a diff

proski:
```
*  Use "patch" extension when editing a patch with a diff
   
   It enables syntax highlighting in many editors. The file names are the
   same as in the Python implementation.
```

See https://github.com/stacked-git/stgit/blob/v1.5/stgit/lib/edit.py#L105 for the Python implementation.

Tested locally by `stg edit` and `stg edit -d`.


jpgrayson:
Thank you also for this PR, @proski.

Issue #203, opened 2022-08-20
stg uncommit does not recognize annotated tags


larsks:
I was working with [a repository](https://git.openldap.org/openldap/openldap) and I wanted to create a patch stack from the tag `OPENLDAP_REL_ENG_2_6_3`, so I did this:

```
git checkout -b 2.6.3-patches OPENLDAP_REL_ENG_2_6_3
stg init
...a series of `git am` commands...
stg uncommit -t OPENLDAP_REL_ENG_2_6_3 -x
```

This failed with:

```
error: Target `OPENLDAP_REL_ENG_2_6_3` is not a commit
```

And indeed, `OPENLDAP_REL_ENG_2_6_3` is an annotated tag:

```
$ git cat-file -t OPENLDAP_REL_ENG_2_6_3
tag
```

This was surprising because I generally expect to be able to use tags -- both simple and annotated -- in place of commit references without problems. The workaround was to spell the tag name like this:

```
stg uncommit -t OPENLDAP_REL_ENG_2_6_3^{commit} -x
```

Is this something that stgit should handle automatically?

PR #204, opened 2022-08-20
Allow uncommit to work with annotated tags

larsks:
Use the git2::Object.peel method to recursively dereference objects
until we find a commit. This allows uncommit to work with things that
wrap commit objects (like annotated tags).

Closes: #203

### DIFF
--- a/src/cmd/uncommit.rs
+++ b/src/cmd/uncommit.rs
@@ -100,9 +100,12 @@ fn run(matches: &ArgMatches) -> Result<()> {
         let target_object = repo
             .revparse_single(commitish)
             .map_err(|_| anyhow!("Invalid commitish `{commitish}`"))?;
+
         let mut target_commit = target_object
-            .into_commit()
-            .map_err(|_| anyhow!("Target `{commitish}` is not a commit"))?;
+            .peel(git2::ObjectType::Commit).ok()
+            .and_then(|c| c.into_commit().ok())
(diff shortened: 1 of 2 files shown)


larsks:
So, fair warning, I know absolutely nothing about Rust, but I wanted to contribute a solution rather than just pointing out a problem. I'd be happy to update this as necessary!

src/cmd/uncommit.rs
```diff
@@ -100,9 +100,13 @@ fn run(matches: &ArgMatches) -> Result<()> {
         let target_object = repo
             .revparse_single(commitish)
             .map_err(|_| anyhow!("Invalid commitish `{commitish}`"))?;
+
         let mut target_commit = target_object
+            .peel(git2::ObjectType::Commit)
+            .map_err(|_| anyhow!("Target `{commitish}` cannot be evaluated as a commit"))?
             .into_commit()
-            .map_err(|_| anyhow!("Target `{commitish}` is not a commit"))?;
+            .unwrap();
```
larsks, line 108: I am assuming here that if the call to `peel` is successful that we don't need to check for errors from `into_commit()`. Does that seem reasonable?

larsks: Yeah, that was ugly. I've changed it.

jpgrayson:
Thanks for this PR, @larsks. I've committed your patch and made a couple follow-up changes. This was a really good first try with Rust!

Issue #195, opened 2022-07-29
[Upstream Issue] Libgit2 and transitively Stacked Git do not support sparse checkouts


NonLogicalDev:
Upstream Issues:
* https://github.com/libgit2/libgit2/issues/6044
* https://github.com/libgit2/libgit2/issues/2263

Adding this issue to track upstream issues of LibGit2 related to sparse checkout support.

Our company is using a huge monorepo, and we have recently onboarded onto Git Sparse checkouts, which seem to be unsupported in LibGit2 which Rust version is using.

Despite my attempts at doing custom builds with some PRs I found online for LibGit2, although I could get the Read Only workflow to succeed like `stg series`.

I could not get the `stg push` to work as it seems like sparse index is throwing LibGit2 way off.

jpgrayson:
Thanks for writing up this issue, @NonLogicalDev.

It would be nice if libgit2 supported sparse checkout natively. That said, it doesn't seem like that project has the will to take on the risk of sparse checkout any time soon.

When reimplementing StGit in Rust, a key design decision was how to interact with the git repository. Initially, I tried orienting almost entirely to libgit2--it is fast and its abstractions are pretty nice. However, as I implemented more and more of StGit, I ran into a bunch of compatibility problems; tests would mostly pass, but various edge cases would fail. Patch application and merge resolution (in various contexts) just don't work correctly.

So where libgit2 didn't get the job done, I reverted to using `git` commands.

Coming back to sparse checkouts, it might be that there is yet another subset of operations that StGit is currently doing with libgit2 that could be converted to `git` commands that would enable StGit to just work with sparse checkouts. I think this is a more likely path to success versus waiting for libgit2 to gain sparse checkout support.

Perhaps the line we can draw for libgit2's use in StGit is to only use it as an object and configuration database. I.e. looking up refs, commits, trees, blobs, and config, but any significant mutations of repository or working tree state should be done by `git` proper. At one point, I had done the work to entirely eliminate libgit2 as a dependency (I wrote an entire replacement `stupid2` library), but the runtime performance was significantly worse than the current hybrid approach.

I can think of a couple of concrete next steps for this issue:
1. Define some sparse checkout test cases that we want to work, covering at least `new`, `refresh`, `push`, and `pop`.
2. Audit for the remaining places where StGit uses libgit2 for checkout and tree comparisons.

arxanas:
[...]

PR #206, opened 2022-09-09
Upgrade to git2 0.15

abhinav:
Since 0.13, git2 has seen [many changes][1], including rust-lang/git2-rs#791
which should address the extensions.worktreeConfig issue in #195.

[1]: https://github.com/rust-lang/git2-rs/compare/0.13.0...0.15.0

The upgrade required fixing uses of [ConfigEntries][2] because it's no
longer an iterator.
However, it provides an API that lends itself well to a `while let` statement,
which provides a for-loop-like experience, so the fix was straightforward.

[2]: https://docs.rs/git2/0.15.0/git2/struct.ConfigEntries.html

### DIFF
--- a/src/stack/stack.rs
+++ b/src/stack/stack.rs
@@ -86,15 +86,17 @@ impl<'repo> Stack<'repo> {
         }
         state_ref.delete()?;
         let mut config_to_delete: Vec<_> = Vec::new();
-        for entry in config
-            .entries(Some(&format!("branch.{branch_name}.stgit")))?
-            .into_iter()
+
         {
-            let entry = entry?;
-            if let Some(name) = entry.name() {
-                config_to_delete.push(name.to_string());
+            let mut entries = config.entries(Some(&format!("branch.{branch_name}.stgit")))?;
+            while let Some(entry) = entries.next() {
+                let entry = entry?;
+                if let Some(name) = entry.name() {
+                    config_to_delete.push(name.to_string());
+                }
             }
         }
+
         for name_to_delete in config_to_delete {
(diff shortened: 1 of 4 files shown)


src/stack/stack.rs
```diff
@@ -86,15 +86,17 @@ impl<'repo> Stack<'repo> {
...
-            if let Some(name) = entry.name() {
-                config_to_delete.push(name.to_string());
+            let mut entries = config.entries(Some(&format!("branch.{branch_name}.stgit")))?;
+            while let Some(entry) = entries.next() {
+                let entry = entry?;
+                if let Some(name) = entry.name() {
+                    config_to_delete.push(name.to_string());
+                }
             }
         }
```
abhinav, line 98: Note, this extra `{ ... }` is required to limit the lifetime of ConfigEntries so that config can be borrowed mutably again below for config.remove.

jpgrayson:
Thanks, @abhinav! I'm a little embarrassed to not have noticed that StGit had gotten so behind on `git2`. I'm going to have up my `cargo outdated` game.

abhinav:
> Thanks, @abhinav! I'm a little embarrassed to not have noticed that StGit had gotten so behind on `git2`. I'm going to have up my `cargo outdated` game.

GitHub's dependabot supports Cargo, if that's something you're willing to use.
You can stick this in .github/dependabot.yml and it'll check for updates and make PRs at your chosen cadence.

```yaml
version: 2
updates:
  - package-ecosystem: "cargo"
    directory: "/"
    schedule:
      interval: "weekly"
```

PR #207, opened 2022-09-10
Set up dependabot

abhinav:
Adds a dependabot configuration to check for updates of dependencies once a week.

Based on comments in https://github.com/stacked-git/stgit/pull/206#issuecomment-1242573201.
Feel free to close if you'd rather update dependencies on your own schedule.


jpgrayson:
Cool. Let's give this a try. Thanks, @abhinav!

PR #216, opened 2022-09-22
Better support patch names starting with hyphen

bb010g:
Also, solve some other `--`-related ambiguities.

This reverts commit 0632851.

---

I personally use patch names starting with `-` a fair amount. The command line syntax here isn't the prettiest, but that's what you get when you start out with positional args before the `--`. This is a problem when any positional arg before the separating `--` needs to be `--`.


jpgrayson:
I appreciate this PR, @bb010g. It's good to know that at least one person wants/uses patch names with leading hyphens.

I am not going to accept this PR, but I have an alternative approach that I think is better.

The alternative is to allow/recognize patch names with leading hyphens if the leading hyphen is escaped with backslash. Some examples:

```sh
$ stg new \\-patch
# Literal backslash may be used inside single quotes
$ stg new '\-patch2'
# Double quotes also require the backslash itself to be escaped
$ stg new "\\-patch3"
# Patch ranges also support this syntax
$ stg series \\-patch..-patch3
# Support commands that take patch names and pathspecs after -- separator
$ stg show \\-patch -- file.txt
# Backslash escaped hyphen is not needed for patch names after the -- separator
$ stg goto -- -patch
# ...but the backslashed hyphen is still allowed after the -- separator
$ stg goto -- \\-patch
# A patch named "--" is also supported
$ stg new \\--
```

I have a patch in my stack that enables the above patch name syntax.

This approach avoids introducing any new command line options. And it avoids having options that have the same meaning as arguments, which is the thing I'm not willing to accept.

Using the backslash to escape a leading hyphen is conventional. And it is sufficiently unambiguous because backslashes are forbidden from patch names (because they are forbidden from git reference names).

jpgrayson:
I've pushed the alternative solution to this problem in 8fc065025295e574f23f0ebfbfbd2590095d4e46.

bb010g:
I don't get why this approach is better for commands like `stg new`. `stg new -- --foo` is standard and unambiguous on the CLI; same with `stg refresh --patch=--foo`. Now the patch names need processing there, and your approach causes problems when you have patches that start with `\-` (a pretty niche case, but still a special case). Dealing with this syntax makes scripting `stg` more complex.

I think it'd be better to change the CLIs of `stg email format` and `stg email send` to always use options and be unambiguous by default, whether that's with `--patch <patch>` arguments or `--git-arg <arg>` like you might see with a C compiler for passing options through.

jpgrayson:
> I don't get why this approach is better for commands like `stg new`. `stg new -- --foo` is standard and unambiguous on the CLI

In StGit 2.0, `stg new` supports immediately refreshing using the `-r`/`--refresh` option. And like `stg refresh`, `stg new --refresh` can take pathspecs after a `--` separator. So in StGit 2.0, without one of these changes it was *impossible* to create a new patch with a leading '-' name. It was only possible to get a `-patch-name` using `stg rename` or `stg edit`. This was the motivation for removing support for leading '-' patch names.

`stg show` in StGit 2.0 has the same problem with the same cause since it also now accepts pathspecs. And thus specifying a `-patch` on the command line was similarly impossible.

For other StGit commands that accept patch name arguments, but not also pathspec arguments, it remains possible to use an unescaped `-patch` name after `--`.

> Dealing with this syntax makes scripting stg more complex.

I think this is a strong point. For the subset of commands where the patch names cannot come after `--`, a correct script would have to recognize patch names with leading '-' and escape them.

> your approach causes problems when you have patches that start with `\-`

That's not an issue because such patch names are (and always have been) forbidden. StGit patch name rules are a superset of Git reference name rules (see git-check-ref-format(1)) and backslashes are not allowed in Git reference names. So when StGit sees `\-patch-name` on the command line, it unambiguously expresses `-patch-name`.

> I think it'd be better to change the CLIs of `stg email format` and `stg email send` to always use options and be unambiguous by default, whether that's with `--patch <patch>` arguments or `--git-arg <arg>` like you might see with a C compiler for passing options through.

For `stg email`, you make another compelling point about having a `--git-arg` option instead of accepting git options after `--`. I would prefer having `--git-arg` versus naming sources with `--patch` because the sources aren't necessarily patch names and specifying sources with options would make `stg email` have different calling conventions than `git format-patch` and `git send-email`.

So I think changing how git args are passed through in `stg email` is a change we should make.

For `stg new`, having an alternative option for specifying the patch name in addition to the normal argument might be the least bad path forward to ease the burden on scripts. I would want the option to be `-n`/`--name` (like `stg squash`). I'd also want to consider the merits of having such an option be hidden (i.e. such that it did not show up in the `stg new -h` output).

For `stg show`, the `-p`/`--patch` option as this PR implements may again be the least bad path forward. I'd also want to strongly consider the pros/cons of hiding this option. Since it would be primarily for scripts, does it need to be advertised to all users?

I think we should keep the `\-patch-name` syntax regardless of any of the above changes.

I really appreciate you pushing back on this @bb010g.

PR #220, opened 2022-10-07
Don't split passthrough options on whitespace

bb010g:
Otherwise, `stg email send -G '--subject=Latest patch'` translates to `git send-mail '--subject=Latest' 'patch'`, and the command errors with
```
error: `git send-email`: fatal: ambiguous argument 'patch': unknown revision or path not in the working tree.
```

Also, Zsh completion reuses the current `git` completion, whatever it is. Ideally, `stg -C foo diff -D --bar -D --baz=` would complete the same as `git -C foo diff --bar --baz=`, but it's currently limited to completing the same as `git -C foo diff --baz=`, so multi-word options won't complete properly (not that they would before). If Zsh provided a companion associative array to `opt_args` that contained indices for each value represented in `opt_args`, we could filter to the current word and provide virtually perfect completion, but this will involve submitting a patch to Zsh to enhance both `comparguments` and `_arguments`.

### DIFF
--- a/completion/stgit.zsh
+++ b/completion/stgit.zsh
@@ -1208,24 +1208,45 @@ __stg_add_args_trailers() {
 }
 
 __stg_git_diff_opts() {
-    local -a diff_opts
-    diff_opts=(${(z)"$(_call_program git-diff-options "git ${__stg_C_args} diff-tree --git-completion-helper")"})
-    __stg_git_command_successful $pipestatus || return 1
-    _wanted git-diff-options expl "diff option" compadd -a diff_opts
+    words=('git' ${(@)__stg_C_args} 'diff-tree')
+    # Needs a way to know word indices for opt_args values to only provide through CURRENT.
+    #
+    # words+=(${(0)opt_args[-O]} ${(0)opt_args[--diff-opts]})
+    words+=("$PREFIX$SUFFIX")
+    (( CURRENT = $#words ))
+
+    local _comp_command1 _comp_command2 _comp_command
+    _set_command
+    _message -e git-diff-option 'diff option' && \
+        _dispatch "$_comp_command" "$_comp_command1" "$_comp_command2" -default-
 }
 
 __stg_git_format_patch_opts() {
-    local -a format_patch_opts
-    format_patch_opts=(${(z)"$(_call_program git-format-patch-options "git ${__stg_C_args} format-patch --git-completion-helper")"})
-    __stg_git_command_successful $pipestatus || return 1
-    _wanted git-format-patch-options expl "format-patch option" compadd -a format_patch_opts
+    words=('git' ${(@)__stg_C_args} 'format-patch')
+    # Needs a way to know word indices for opt_args values to only provide through CURRENT.
+    #
+    # words+=(${(0)opt_args[-G]} ${(0)opt_args[--git-opts]})
+    words+=("$PREFIX$SUFFIX")
+    (( CURRENT = $#words ))
+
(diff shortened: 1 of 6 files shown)


jpgrayson:
Thank you for this PR, @bb010g.

I am in agreement that for the `--git-opts` options to `stg email send` and `stg email format` it will be better to support a single git option per StGit `--git-opts` occurrence. This is a good use case to accommodate correctly.

For diff options, the Python implementation allows multiple space-separated options as the value to `--diff-opts` as well as allowing multiple occurrences of `--diff-opts`, which is why this was implemented the way it was. This relates to #187.

By only allowing a single git diff option per `--diff-opts`, there is a compatibility break with StGit 1.x. I am not sure how attached anyone might be to being able to specify multiple space-separated options, which has made me hesitant to make this change. That said, I agree that a 1:1 mapping between a `--diff-opts` occurrence and a git diff option is conceptually cleaner and solves the problem of supporting values with spaces.

A question I had was whether git diff options were exposed to the same problem you identified with `git email send -G '--subject=Latest patch'`? I.e. are there any `git diff-tree` options that may need space-separated values? Here is what I'm seeing:

- `--anchored=<text>`
- `--word-diff-regex=<regex>`
- `--color-words[=<regex>]`
- `-S<string>`
- `-G<regex>`
- `--relative[=<path>]`
- `-I<regex>`, `--ignore-matching-lines=<regex>`
- `--line-prefix=<prefix>`
- `--pretty[=<format>]`

So yes, a lot of options that where a user may want a space separated value. I'm in agreement that git diff options should also support values with spaces in them. The most straightforward way of doing this is the change you've made to have the 1:1 mapping between StGit `-O`/`--diff-opts` occurrences and git diff options. I can imagine other, less straightforward approaches that involve StGit having its own shell-like quoting rules, but I think that path, in addition to being much more complicated to implement, will be worse for users than this small compatibility break with StGit 1.x.

So I am inclined to merge this change.

However, I am also going to do some follow-on changes:
- Rename `--diff-opts` to `--diff-opt`. I will retain `--diff-opts` as a hidden alias.
- Rename `--git-opts` to `--git-opt`. No alias for this one since it is new to StGit 2.0.
- Repair the now broken tests.
- Add tests for `--diff-opt` options with value containing spaces. E.g. `--diff-opt=--line-prefix="x x x "`.
- Repair code formatting.
- Improve the zsh completions to:
  - Comprehend already-specified `--diff-opt` and `--git-opt` values. I.e. so that already used options are excluded from the completions.
  - Only offer completions for git options and not arguments to `git diff-tree`/`git send-email`/`git format-patch`.

bb010g:
Thank you!

If you're looking at making those sorts of changes, renaming `--diff-opt` to `--diff-tree-opt` for clarity might be a good idea, or at least referring to it in documentation as `<diff-tree-opt>`.

> Improve the zsh completions to:
> 
>   * Comprehend already-specified `--diff-opt` and `--git-opt` values. I.e. so that already used options are excluded from the completions.
> 
>   * Only offer completions for git options and not arguments to `git diff-tree`/`git send-email`/`git format-patch`.

From what I discovered while writing this patch, I expect both of those to be rather challenging. For the former, you want to accumulate `--diff-opt` values up to that point using the same parsing strategy as `_arguments`, but without actually using `_arguments` (which is hard! `_arguments` parses via `comparguments`, which is a C-coded Zsh builtin), and for the latter, you want to somehow inform the behavior of a completion function that you have no knowledge of. These are both points that, to my knowledge, Zsh itself needs improvements on. All that said, best of luck solving these problems.

PR #223, opened 2022-10-11
Zsh refine completion for passthrough options

bb010g:
Ensure `zparseopts` is loaded from `zsh/zutil`.

Pass through options after the `CURRENT` option.

Use shell builtins (not `seq`) & optimize.

---

Nice job with figuring out a decent solution. I think `zparseopts` could have issues if other keyed options took values starting with hyphen, but it should work fine until then.

I wish Zsh allowed subscripting like `${git_opts[2,2,$#git_opts-2]}`, but it sadly doesn't.

### DIFF
--- a/completion/stgit.zsh
+++ b/completion/stgit.zsh
@@ -1207,38 +1207,42 @@ __stg_add_args_trailers() {
 }
 
 __stg_complete_git_opts() {
-    local gitcmd short long
-    gitcmd=$1
+    local git_cmd short long i
+    git_cmd=$1
     short=$2
     long=$3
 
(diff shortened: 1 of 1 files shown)


jpgrayson:
I appreciate the collaboration on this, @bb010g. It's nice to have someone else looking at the zsh code.

> Nice job with figuring out a decent solution. I think zparseopts could have issues if other keyed options took values starting with hyphen, but it should work fine until then.

Thanks. Agreed it is not bulletproof, but will likely be good enough.

> I wish Zsh allowed subscripting like ${git_opts[2,2,$#git_opts-2]}, but it sadly doesn't.

I had the same thought.

PR #236, opened 2022-11-20
fix: add support for git version parsing on MacOS.

Byron:
Such version strings are special as they look like `2.37.1 (Apple Git-137.1)`, and the parsing was adjusted to account for that.


jpgrayson:
Thank you!

PR #245, opened 2022-12-15
Add shorthand -s for --signoff for the edit command

liuw:
Stgit's email format command and git-commit(1) already support this shorthand.

Signed-off-by: Wei Liu <liuw@liuw.name>

### DIFF
--- a/src/patchedit/mod.rs
+++ b/src/patchedit/mod.rs
@@ -128,6 +128,7 @@ pub(crate) fn add_args(
         .arg(
             Arg::new("signoff")
                 .long("signoff")
+                .short('s')
                 .alias("sign")
                 .help("Add Signed-off-by message trailer")
                 .long_help(


jpgrayson:
Thank you for this PR, @liuw.

The proposed `-s` short option for `--signoff` conflicts in `stg new` with `--submodules`/`-s`. This will trigger a panic at runtime.

I would be okay with removing `-s` from `--submodules`, but this would be a breaking change and could mess up some people's workflows. This means that we would at least want to wait until the next minor release (v2.2.0) before introducing this change.

liuw:
> Thank you for this PR, @liuw.
> 
> The proposed `-s` short option for `--signoff` conflicts in `stg new` with `--submodules`/`-s`. This will trigger a panic at runtime.
> 
> I would be okay with removing `-s` from `--submodules`, but this would be a breaking change and could mess up some people's workflows. This means that we would at least want to wait until the next minor release (v2.2.0) before introducing this change.

Interesting. I didn't know that. I just started exploring using stgit for my workflow these two days. I myself don't use submodule so I didn't hit the conflict.

Feel free to do whatever to this PR. Missing `-s` is not a deal breaker for me.

jpgrayson:
I'm going to leave this PR open as a reminder to consider this change as we get closer to the next minor release.

I appreciate the contribution.

And I hope that StGit is proving helpful for your workflow!

jpgrayson:
I've made this change in 2126645. Thanks for motivating this change, @liuw.

PR #266, opened 2023-01-30
Use correct base directory for core.hooksPaths

martijnthe:
In the .py => .rs rewrite, the determination of the base directory changed a bit. Before it was using the current working directory. After, it was using the .git directory instead. Both are somewhat incorrect.

According to [this
documentation](https://git-scm.com/docs/githooks/2.9.5#_description), the .git directory should be used in case the repo is a bare one. In case, the repo is a non-bare one, the root of the working tree should be used as base directory for the `core.hooksPath`.

The existing test cases (t2703-refresh-pre-commit-hook.sh) did not catch any of this, mainly because the tests did not assert whether the hooks script was actually being run at all.

This patch fixes this.

### DIFF
--- a/src/hook.rs
+++ b/src/hook.rs
@@ -4,7 +4,7 @@
 
 use std::{
     io::Write,
-    path::{Path, PathBuf},
+    path::PathBuf,
 };
 
 use anyhow::{anyhow, Context, Result};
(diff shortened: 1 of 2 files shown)


jpgrayson:
Thank you for this PR @martijnthe. Much appreciated.

I'm going to merge this PR and follow-up with a refactor commit to handle the lint issues.

PR #277, opened 2023-02-17
fix: enable local-time in commits.

Byron:
Previously it was necessary to build with special RUSTC_FLAGS to turn on local offset support. Now the `time` crate makes this a runtime capability that each application has to opt-in to.

----

I usually build my binaries locally and at some point must have built without activating local offsets. Since `time` v0.3.18 has a more obvious mechanism for it this PR is meant to add it to `stacked-git`.

Further, I thought I'd update to the latest `gix` for good measure, even though that isn't necessary to get local times to work.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -245,6 +245,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "dunce"
+version = "1.0.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


jpgrayson:
I appreciate this PR, @Byron. Thank you!

Byron:
You are very welcome! Please also consider to upgrade to v0.36.1 to not see breakage 😅.

PR #307, opened 2023-04-01
chore: update gix-revision, fixes #306

### LINKED ISSUE #306: `stg rebase <ref>` does not work when ref contains @
Hi everyone.

I updated to the lastest version today, and found a regression related to reference handling. As far as I can tell, git revisions are not recognized when they contain certain character. At `$WORK` we use @ in our branch names, for instance.

How to reproduce

```
git init
touch t
git add t
git commit -m "add t"
git branch lthms@reproduce
stg init
stg rebase lthms@reproduce
```

was working with 2.1.0, but does not work with 2.2.1.

lthms:
Fixes https://github.com/stacked-git/stgit/issues/306

The underlying cause of the error reported in the linked issue was a bug in `gitoxide` which got fixed in the latest release.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -763,9 +763,9 @@ dependencies = [
 
 [[package]]
 name = "gix-revision"
-version = "0.12.1"
+version = "0.12.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


jpgrayson:
Thank you so much for working this issue upstream, @lthms. And thanks to @byron for his continued support of StGit.

lthms:
That was fast :rocket: and thanks for the release!

Issue #235, opened 2022-11-16
How update old branches that give error: Malformed metadata (expected version 4)


jkleckner:
I have some old branches that give the error: `error: Malformed metadata (expected version 4)`

What would be the best way to update them to work with the new stgit?

jpgrayson:
StGit 2.0 is only able to upgrade from stack metadata version >= 4, which was introduced in StGit 1.0. So your old branches were ostensibly initialized using StGit <1.0.

The easiest path forward is to upgrade these branches using StGit 1.5. Stack upgrade can be triggered for all branches within a repository by running `stg branch --list`.

You can run StGit 1.5 pretty easily by using a clone of the stgit repository. Here are the complete steps:

```shellsession
$ git clone --depth=1 --branch v1.5 https://github.com/stacked-git/stgit.git path/to/stgit-1.5
$ PYTHONPATH=path/to/stgit-1.5 python -m stgit version
$ cd path/to/repo-to-upgrade
$ PYTHONPATH=path/to/stgit-1.5 python -m stgit branch --list
$ rm -rf path/to/stgit-1.5
```
---

N.B. it would be possible add code to StGit 2.x to be able to upgrade from these older stack metadata formats, but I wanted to limit the scope of the Rust implementation and I thought compatibility with StGit 1.x would be sufficient. I'm hoping the above upgrade path will be sufficient for users jumping directly from StGit <1.0 to StGit 2.x.

jkleckner:
@jpgrayson Thank you that worked perfectly!  I ran into this when cleaning up some very old branches.

AndreRH:
Since #318 was merged this automagically works now :)

PR #318, opened 2023-05-07
Upgrade from older stack state formats

AndreRH:
Hi,

As Debian derivates are stuck pre stgit 1.0 for years now, many people including me likely have a huge amount of version 2 stgit branches around. While there's a workaround as described in https://github.com/stacked-git/stgit/issues/235 , I think it would be more convenient to have support for upgrading from at least v2 and v3.
I make this a draft PR to gather some initial feedback before I'm done developing it, so please share your opinion and comment on the code.


jpgrayson:
Hi @AndreRH. Thanks for taking this on. It would be nice for StGit 2.x to be capable of performing these stack upgrades.

The code pattern in this PR looks fine to me. Carry on!

You will probably want to resurrect `t4000-upgrade.sh` from the last Python release to re-enable testing of upgrades from these older stack versions.

AndreRH:
Yeah, I like Rust, OSS and stgit. So this was the missing feature I was waiting for to do my first OSS contribution in Rust :)
Thanks for the hint regarding the tests, that was very straight forward.
I removed the draft flag, so be free to review the current state, after all it wasn't that hard.

jpgrayson:
Thank you very much for this contribution, @AndreRH!

PR #322, opened 2023-05-19
fix: importing patch series exported using numbered order will not remove the numbers

aelgogary:
Hi,

Here is a problem i discovered while working with stgit.

if patch series was exported using numbered order `stg export -n`, importing the series `stg import -S -t patches/series` will keep the patch number as part of its name, reexporting previously imported patches than will add the numbers again e.g. "01-01-patch-name.patch"

Thanks

### DIFF
--- a/src/cmd/import.rs
+++ b/src/cmd/import.rs
@@ -703,8 +703,8 @@ fn create_patch<'repo>(
 }
 
 fn stripname(name: &str) -> &str {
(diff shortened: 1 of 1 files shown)


jpgrayson:
Thanks for finding and fixing this issue, @aelgogary. I fixed the formatting issue in your commit and pushed it in c513d2f42478dd055af0d66bbde86c60223b5877. I also did a follow-on change to only strip either a ".diff" or a ".patch" suffix, but not both.

PR #350, opened 2023-07-21
fix(import): Keep first line break in body

fmaurer-rh:
When importing a patch with multiple lines, the first line break in the body was removed from the body of the commit in the stack, because it was stripped when searching for the headers. Add the line break again when adding the first line to the split_message.

This did not surface in the tests, because the test bodies always had one line and the expected bodies had an incorrect number of line breaks at the end. Fix the expected number of line breaks as well.


jpgrayson:
Thank you for finding and fixing this issue, @fmaurer-rh. Much appreciated.

PR #416, opened 2024-02-09
fix: Pass stdio for interactive editing

### LINKED ISSUE #415: Interactive editor (vim) broken on 2.4.3
After updating stgit from 2.4.2 to 2.4.3 I observe the following behavior on macOS: when I open an interactive editor (vim in my case), it complains that the input is not a terminal. Afterwards, the terminal is broken and can only be fixed by running `reset` (note the additional indentation after `stg edit`, the calls to `stg series` are only to visualize the issue):

```
user@host$ stg series
+ change-foo
> add-bar
user@host$ stg edit  
hint: Waiting for your editor to close the file... Vim: Warning: Input is not from a terminal

& add-bar
         > add-bar
                  % 
user@host$ stg series
+ change-foo
            > add-bar
                     %
user@host$
```

When I do the same inside tmux, it gets even worse: inside vim, many keys do not work as expected, e.g., typing backspace produces `^?` instead. 

Environment info:
- macOS
- stgit 2.4.3
- git 2.43.0
- zsh 5.9
- tmux 3.3a
- vim 9.1

Bisecting between v2.4.2 and v2.4.3 shows that 048f182a ("fix: use gix-command for interactive edit") is the commit that introduced the bug. I assume that gix_command somehow messes with the stdin.

skyfmmf:
gix_command by default does not pass stdin to commands. This caused editors and shells to behave weirdly. For example, vim printed a warning during opening:

Vim: Warning: Input is not from a terminal

After closing vim started this way, the terminal was messed up and needed an reset to print outputs correctly.

This is an attempt to fix stacked-git/stgit#415. It fixes the issue described there, but I can not say how it will affect other editors or even scripts. It may re-introduce some errors that 048f182a ("fix: use gix-command for interactive edit") originally tried to fix.

Closes #415

### DIFF
--- a/src/patch/edit/interactive.rs
+++ b/src/patch/edit/interactive.rs
@@ -92,6 +92,7 @@ pub(crate) fn call_editor<P: AsRef<Path>>(
         let status_result = gix_command::prepare(&editor)
             .arg(path.as_ref())
             .with_shell_allow_argument_splitting()
+            .stdin(std::process::Stdio::inherit())
             .stdout(std::process::Stdio::inherit())
             .spawn()
             .with_context(|| format!("running editor: {}", editor.to_string_lossy()))


jpgrayson:
Thanks for this PR, @skyfmmf. On which platform(s) did you observe this behavior?

skyfmmf:
> On which platform(s) did you observe this behavior?

I'm seeing this with this environment:
- macOS 14.3
- stgit 2.4.3
- git 2.43.0
- zsh 5.9
- tmux 3.3a
- vim 9.1

jpgrayson:
Thanks. I have a similar environment, but using neovim instead of vim. I can reproduce the behavior with vim, but do not see it with neovim or emacsclient.

PR #437, opened 2024-04-05
Update gix, gix-command and clap to the latest versions

blinxen:
I am currently updating `stgit` in Fedora to the latest version and saw that you should be able to safely update `gix` and `clap` in `Cargo.toml`.

I concluded that it is safe, since it can be built and all tests pass. See https://download.copr.fedorainfracloud.org/results/blinxen/gitoxide/fedora-rawhide-x86_64/07262346-stgit/builder-live.log.gz

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,6 +9,24 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
+name = "ahash"
+version = "0.8.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e89da841a80418a9b391ebaea17f5c112ffaaa96f621d2c285b5174da76b9011"
+dependencies = [
+ "cfg-if",
+ "once_cell",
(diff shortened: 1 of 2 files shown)


jpgrayson:
Thanks for the PR. I've updated gix to 0.61.1 in 0d9c9f7978c18e1029af9bfbb96014146bce2513.

blinxen:
Thanks! Don't forget to also update `gix-command`.

jpgrayson:
I did not update `gix-command` from 0.3.4 to 0.3.6 in `Cargo.toml`, but it is updated to 0.3.6 in `Cargo.lock` since 0.3.6 is semver compatible with 0.3.4.

For purposes of Fedora's packaging, is the distinction between `gix-command` 0.3.4 and 0.3.6 in the `Cargo.toml` file material?

blinxen:
> I did not update gix-command from 0.3.4 to 0.3.6 in Cargo.toml, but it is updated to 0.3.6 in Cargo.lock since 0.3.6 is semver compatible with 0.3.4.

Ah, you are right. Totally missed that!

> For purposes of Fedora's packaging, is the distinction between gix-command 0.3.4 and 0.3.6 in the Cargo.toml file material?

No

jpgrayson:
Great! Thanks again!

PR #443, opened 2024-04-15
Update gix to version 0.62

blinxen:
This update contains a security fix for
https://rustsec.org/advisories/RUSTSEC-2024-0335.html

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -323,9 +323,9 @@ dependencies = [
 
 [[package]]
 name = "gix"
-version = "0.61.0"
+version = "0.62.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


jpgrayson:
Thanks for the PR, @blinxen.

I'll note that StGit does not use the affected gix-transport API directly, or AFAIK indirectly.

StGit does not actually do any transport operations anymore (`stg clone` used to be a command, but was removed a while back).

So my off the cuff opinion is that there is no exploit vector for this via StGit.

blinxen:
That is true, however this change is motivated by the gitoxide update in Fedora. Because the project (gitoxide) consists of around 70 crates, we can't support multiple versions of it. That is why I would like to update all Fedora packages that require gitoxide to use the new version. This makes the life of package maintainers easier :D.

blinxen:
I would appreciate it if this could get merged however it doesn't really have a big priority right now.

PR #466, opened 2024-06-13
fix: prefer --authdate over patch time with author

### LINKED ISSUE #457: Question about stg edit
I have a couple aliases I use as part of our backporting work:

```
	sd = edit --authdate "now"
	ra = edit --signoff --author 'Jerry Snitselaar <jsnitsel@redhat.com>' -e
```

If I try to just put the `--authdate "now"` in the `ra` alias it will not adjust the date. It only seems to work when it is by itself. Is that intentional? It isn't a big deal, but just something I'd noticed.

kmerfeld:
fixes #457


jpgrayson:
I do not recall a reason why the commit date should be preferred over a `--authdate` command line option. This change looks like a sensible correction to me. Thank you!

Issue #471, opened 2024-07-15
"stg import --reject" is broken


dtor:
According to the documentation "stg import --reject" is supposed to add an empty patch to the stack. It also says that failed diff will be saved as .stgit-failed which does not make much sense and is different from how old versions of stg behaved (i.e. apply what it can to the working tree and leave the rest as *.rej files). It also contradicts explanation to the "--reject" option which says that it will "leave rejected hunks in .rej files".

Unfortunately the current behavior is to partially apply to the work tree and leave .rej (which is good), but no commit is created.

PR #472, opened 2024-07-15
fix: stg import --reject should create empty commit

dtor:
When trying to import a patch that does not apply cleanly, and using --reject option, stg should apply what it can, leave the rest in .rej files, and create an empty commit. The work to apply the patch is outsourced to "git apply --reject" which exits with status code 1 if patch is applied partially and it is treated as error by stg import.

Fix the issue by treating exit codes less than 128 from "git apply --reject" as success (128 signals hard error).

Also add a test case and fix up documentation for "stg import".

Closes: #471

### DIFF
--- a/src/cmd/import.rs
+++ b/src/cmd/import.rs
@@ -44,8 +44,10 @@ fn make() -> clap::Command {
              allows the patches source to be fetched from a url instead of from a \
              local file.\n\
              \n\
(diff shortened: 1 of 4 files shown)


dtor:
This does not agree with "stg fold --reject", I will have to rework this.

PR #474, opened 2024-07-15
fix: harden "stg rebase -i" against patches with milti-line subjects

dtor:
I have a stgit stack that is very old (I started it more than 10 years ago), so it has all kinds of WIP patches that I started and then did not quite finish. In several of them I have malformed patch subjects, such as multi-line titles:

$ git show -1 --pretty=%B
subsystem: fix driver
certain change

This breaks "stg rebase -i" because it expects to have one line per patch in the rebase template, but in this case subject spills over to the next line which rebase mistakes for an instruction:

error: unknown instruction action `certain`

Fix this my replacing new lines and line feeds in the subject with spaces.


jpgrayson:
Thank you for this PR, @dtor. Much appreciated.

PR #477, opened 2024-07-28
fix: stg import --reject should create empty commit

dtor:
When trying to import a patch that does not apply cleanly, and using --reject option, stg should apply what it can, leave the rest in .rej files, and create an empty commit. The work to apply the patch is outsourced to "git apply --reject" which exits with status code 1 if patch is applied partially and it is treated as error by stg import.

Fix the issue by not treating return code of 1 from "git apply" as an error when "--reject" option is specified, but rather saving its output, printing it for the user, and continuing with the rest of the import logic. Exit with CONFLICT_ERROR rather than COMMAND_ERROR when the patch does not import/apply cleanly.

"stg fold" reuses much of the same code so it has to be adjusted in the similar fashion. It will also exit with CONFLICT_ERROR when a patch does not apply cleanly.

Also add a test case and fix up documentation for "stg import" and adjust test case for "stg fold".

Closes: #471

### DIFF
--- a/src/cmd/fold.rs
+++ b/src/cmd/fold.rs
@@ -11,6 +11,7 @@ use clap::{Arg, ArgGroup};
 use crate::{
     ext::{CommitExtended, RepositoryExtended},
     patch::SingleRevisionSpec,
+    print_info_message,
     stack::{InitializationPolicy, Stack, StackAccess, StackStateAccess},
     stupid::Stupid,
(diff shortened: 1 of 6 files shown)


jpgrayson:
Thank you for this fix @dtor. It is very much appreciated. Apologies for taking so long to look at this PR.

PR #507, opened 2024-12-20
support calling stg rebase without arguments

fbenkstein:
When calling "git rebase" without a target commit argument then it will try to look up remote tracking branch and rebase onto that. This change makes "stg rebase" behave the same way. Note that this introduces a slight discrepancy with "stg rebase --interactive": in this case it will still use the stack base instead of the remote tracking branch.


fbenkstein:
Sorry for the churn, I was still figuring out the testing.

jpgrayson:
Thanks for this PR. This seems like a sensible change.

PR #536, opened 2025-02-11
stgit.el improvements

MorganJamesSmith:
Hello,

I'm loving stgit!  I'm having lots of fun with it!

stgit.el?  I'm having some troubles.

log-edit was pre-pending "Summary: " when I was trying to edit patches which was frustrating.

It looks very much like the builtin emacs vc but the keybinds are different so that's confusing.

I think I hit a couple of bugs but that might have just been user error.  I decided to try to clean up the code before trying to use it again.

I hope this is helpful!

Thanks,

Morgan

### DIFF
--- a/contrib/stgit.el
+++ b/contrib/stgit.el
@@ -236,65 +236,65 @@ format characters are recognized:
 
 (defface stgit-branch-name-face
   '((t :inherit bold))
-  "The face used for the StGit branch name"
+  "The face used for the StGit branch name."
   :group 'stgit-faces)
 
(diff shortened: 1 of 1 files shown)


jpgrayson:
Thank you for this PR @MorganJamesSmith. As you must have noticed, stgit.el has not seen any substantial maintenance in a long time, so it's great that you've put this effort into it. Much appreciated.

All the changes look sensible to me. Thanks for organizing the commits as you did; made it easy to review.

PR #541, opened 2025-02-17
build: statically linked packages for riscv64

AndreRH:
Now that the riscv64gc musl target is promoted to tier2 and everything is in place, let's build for riscv64
Debian package tested on real HW.


jpgrayson:
Love it. Thank you @AndreRH.

PR #552, opened 2025-04-02
stgit.el improvements #2

MorganJamesSmith:
Hello!  Back again with another round of improvements!

A notable improvement is "Allow empty descriptions" as that was causing spurious and frustrating errors while trying to use it.

Admittedly most of this (6 out of the 8 commits) is just satiating the compiler and should hopefully not cause any changes whatsoever.

Locally testing these changes seems to work but due to a lack of a test suite and the fact that I don't personally use every available code path here, there is a decent chance I broke something somewhere.

### DIFF
--- a/contrib/stgit.el
+++ b/contrib/stgit.el
@@ -59,6 +59,21 @@
 (require 'easymenu)
 (require 'format-spec)
 
+(defvar-local stgit-marked-patches nil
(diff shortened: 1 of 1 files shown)


jpgrayson:
This is really great. Thank you for this. Sorry about the delay getting it merged.

PR #573, opened 2025-06-06
stgit.el improvements #3

### LINKED ISSUE #560: wrong usage of `defvar-local`
```
defvar-local is a Lisp macro in ‘subr.el’.

(defvar-local VAR VAL &optional DOCSTRING)

Define VAR as a buffer-local variable with default value VAL.
Like ‘defvar’ but additionally marks the variable as being automatically
buffer-local wherever it is set.

  Probably introduced at or before Emacs version 24.3.
```

`defvar-local` needs at least two arguments.  However, [only one argument][1] is provided.

related: #552

cc @MorganJamesSmith 

[1]: https://github.com/stacked-git/stgit/blob/503c919dd909e8060fd4de017372f22dde57ac87/contrib/stgit.el#L72-L86

MorganJamesSmith:
Hello again!

This is a pretty small pull request.  Most of the changes are minor documentation changes.

I briefly tested this on Emacs 29.4 so this should fix #560

### DIFF
--- a/contrib/stgit.el
+++ b/contrib/stgit.el
@@ -79,8 +79,7 @@
 (defvar-local stgit-buffer nil)
 
 (defun stgit-set-default (symbol value)
-  "Set default value of SYMBOL to VALUE using `set-default' and
-reload all StGit buffers."
+  "Set default value of SYMBOL to VALUE and reload StGit buffers."
   (set-default symbol value)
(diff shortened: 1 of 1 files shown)


jpgrayson:
Thank you, @MorganJamesSmith!

PR #579, opened 2025-07-20
chore: update gix to 0.73

blinxen:
This PR tries to update `gix` to version 0.73.

The changes were needed because of:

* [`SignatureRef.time`](https://docs.rs/gix-actor/0.35.2/gix_actor/struct.SignatureRef.html#method.time) method now returns a `&str`. There is a [`SignatureRef.time()`](https://docs.rs/gix-actor/0.35.2/gix_actor/struct.SignatureRef.html#method.time) to get [`Time`](https://docs.rs/gix-date/0.10.3/gix_date/struct.Time.html) back. This can however fail and return an Error. I replaced all instances where `time` was accessed directly with `time()?`. In one case, I couldn't use the `?` operator because the method did not return a `Result`. How do you want this to be handled? Is there some default value that can be used here? The compile error looks like this:

```
> cargo b
   Compiling stgit v2.5.3 (/home/hka/dev/opensource/stgit)
error[E0308]: mismatched types
  --> src/ext/signature.rs:49:24
   |
49 |             .unwrap_or(self.time);
   |              --------- ^^^^^^^^^ expected `Time`, found `&str`
   |              |
   |              arguments to this method are incorrect
   |
help: the return type of this call is `&str` due to the type of the argument passed
  --> src/ext/signature.rs:46:20
   |
46 |           let time = matches
   |  ____________________^
47 | |             .get_one::<gix::date::Time>("authdate")
48 | |             .copied()
49 | |             .unwrap_or(self.time);
   | |________________________---------^
   |                          |
   |                          this argument influences the return type of `unwrap_or`
note: method defined here
  --> /builddir/build/BUILD/rust-1.88.0-build/rustc-1.88.0-src/library/core/src/option.rs:1023:12
```

* [`SignatureRef.to_owned()`](https://docs.rs/gix-actor/0.35.2/gix_actor/struct.SignatureRef.html#method.to_owned) can now fail and needs to handled accordingly (by using the `?` operator).
* [`RepositoryExtended.commit_ex()`](https://github.com/stacked-git/stgit/blob/master/src/ext/repository.rs#L56) wants a `SignatureRef` but `Signature` was passed. I assume that because of the updated somewhere `Signature` is returned instead of `SignatureRef`. I added a couple of `Signature.to_ref` calls where needed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,18 +9,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
-name = "ahash"
-version = "0.8.12"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5a15f179cd60c4584b8a8c596927aadc462e27f2ca70c04e0071964a73ba7a75"
-dependencies = [
- "cfg-if",
- "once_cell",
- "version_check",
- "zerocopy",
-]
-
-[[package]]
 name = "allocator-api2"
 version = "0.2.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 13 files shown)


jpgrayson:
Update: I've got a patch that accommodate the breaking changes in gix 0.73. The bad news is that there is a change in gix::date::parse() that causes a regression in StGit.

I've posted an issue upstream: https://github.com/GitoxideLabs/gitoxide/issues/2095

It is possible to work around the regression by not using gix for date parsing, but I'm hoping we can get an upstream fix.

blinxen:
Thanks for looking into this! The gitoxide maintainer released a new version which I hope fixes the issue you mentioned.

jpgrayson:
This is done as of 41aeda52eb2742dcafbf0867eeaae8f972b6f3c1.

The needed fixes went into gix-date 0.10.5.

PR #602, opened 2025-11-20
feat: write .stgit-edit.txt to .git directory instead of current working directory

NonLogicalDev:
Follow git's convention of keeping interactive edit files in the `.git` directory. 
Avoid writing .stgit-edit.txt files to the working tree.

### DIFF
--- a/src/ext/repository.rs
+++ b/src/ext/repository.rs
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: GPL-2.0-only
 
 use std::borrow::Cow;
+use std::path::PathBuf;
 
 use anyhow::{anyhow, Result};
 use bstr::BStr;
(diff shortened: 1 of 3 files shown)


src/ext/repository.rs
```diff
@@ -330,4 +338,12 @@ impl RepositoryExtended for gix::Repository {
                 }
             })
     }
+
+    fn git_data_file(&self, path: &str) -> PathBuf {
+        // If STG_EDIT_IN_CWD is set, return path as is.
+        match std::env::var("STG_EDIT_IN_CWD") {
```
jpgrayson: Is there known cases where someone might need to fallback to the old behavior?

If not, I'd prefer we not have this environment variable at all; keep it simple.

Another issue with this env var is that we're only checking its existence, not its value. So if someone set `STG_EDIT_IN_CWD=0`, for example, it would fallback to the old behavior the same as if they'd set `STG_EDIT_IN_CWD=1`.

PR #627, opened 2026-04-29
update to gix 0.83

Byron:
This is good for various security fixes and performance improvements.


Byron:
🤦‍♂️ So I wasn't up-to-date locally after all!

PR #634, opened 2026-05-31
stgit.el: Fix Index/Work Tree incremental refresh

### LINKED ISSUE #633: stgit.el: Index/Work Tree incremental refresh broken in center mode after lexical-binding change
`stgit.el` no longer refreshes the Index/Work Tree section after state-changing commands such as `stgit-toggle-index` (`i`) when `stgit-show-worktree-mode` is `center` and the stack has unapplied patches.

## Symptoms

In an `stgit-mode` buffer:

1. Have at least one unapplied patch.
2. Use the default/center worktree placement.
3. Put a modified file in the Work Tree section.
4. Press `i` on the file to move it to the index.

Actual behavior:

- Git state changes correctly.
- The `*stgit*` buffer still shows the file in Work Tree.
- Pressing `g` performs a full reload and then the buffer shows the correct state.

Expected behavior:

- The Index and Work Tree sections should refresh immediately after `i`, as they used to.

## Cause

This appears to be a regression from commit 04559f055298df4485116640a3ea76b8e860a0f5 (`stgit.el: Enable lexical-binding`).

When `stgit-show-worktree-mode` is `center`, `stgit-run-series-insert-index` may be called while parsing `stg series` output inside `with-temp-buffer`. Since `stgit-index-node` and `stgit-worktree-node` are buffer-local, the current implementation stores them in the temporary buffer, not the real `stgit-mode` buffer.

As a result, after reload in this layout:

```elisp
stgit-index-node    ;; nil in the stgit buffer
stgit-worktree-node ;; nil in the stgit buffer
```

Then commands like stgit-toggle-index call:

```elisp
(stgit-refresh-worktree)
(stgit-refresh-index)
```

but both are no-ops because the node variables are nil. A full stgit-reload still works, which is why pressing g fixes the display.

## Local workaround

This override fixes the issue locally:

```elisp
;; local patch on upstream stgit regression that does not update the worktree view
;; after git commands (e.g., 'i') in the stgit-show-worktree-mode center.
;; 'bottom' mode is not affected by the bug.
(with-eval-after-load 'stgit
  (defun stgit-run-series-insert-index (ewoc)
    "Insert Index and Work Tree nodes and remember them in the stgit buffer."
    (let ((index-node
           (ewoc-enter-last ewoc
                            (make-stgit-patch
                             :status 'index
                             :name :index
                             :desc nil
                             :empty nil)))
          (worktree-node
           (ewoc-enter-last ewoc
                            (make-stgit-patch
                             :status 'work
                             :name :work
                             :desc nil
                             :empty nil))))
      (with-current-buffer (ewoc--buffer ewoc)
        (setq stgit-index-node (cons ewoc index-node)
              stgit-worktree-node (cons ewoc worktree-node))))))
```

This preserves center mode and restores immediate refresh after i.

jrajahalme:
Index/Work Tree incremental refresh has been broken in center mode after lexical-binding change made in commit 04559f055298df4485116640a3ea76b8e860a0f5.

When stgit-show-worktree-mode is center, stgit-run-series-insert-index may be called while parsing stg series output inside with-temp-buffer. Since stgit-index-node and stgit-worktree-node are buffer-local, the current implementation stores them in the temporary buffer, not the real stgit-mode buffer.

As a result, after reload in this layout:

stgit-index-node    ;; nil in the stgit buffer
stgit-worktree-node ;; nil in the stgit buffer

Then commands like stgit-toggle-index call:

(stgit-refresh-worktree)
(stgit-refresh-index)

but both are no-ops because the node variables are nil. A full stgit-reload still works, which is why pressing g fixes the display.

Fix by applying the updated stgit-index-node and stgit-worktree-node to the real stgit-mode buffer instead.

Fixes: 04559f055298df4485116640a3ea76b8e860a0f5
Fixes: #633

### DIFF
--- a/contrib/stgit.el
+++ b/contrib/stgit.el
@@ -573,18 +573,21 @@ been advised to update the stgit status when necessary.")
     (ewoc-invalidate (car stgit-worktree-node) (cdr stgit-worktree-node))))
 
 (defun stgit-run-series-insert-index (ewoc)
-  (setq stgit-index-node    (cons ewoc (ewoc-enter-last ewoc
-                                                        (make-stgit-patch
-                                                         :status 'index
-                                                         :name :index
-                                                         :desc nil
-                                                         :empty nil)))
-        stgit-worktree-node (cons ewoc (ewoc-enter-last ewoc
-                                                        (make-stgit-patch
-                                                         :status 'work
-                                                         :name :work
-                                                         :desc nil
-                                                         :empty nil)))))
+  (let ((index-node (ewoc-enter-last ewoc
+                                     (make-stgit-patch
+                                      :status 'index
+                                      :name :index
+                                      :desc nil
+                                      :empty nil)))
+        (worktree-node (ewoc-enter-last ewoc
+                                        (make-stgit-patch
+                                         :status 'work
+                                         :name :work
+                                         :desc nil
+                                         :empty nil))))
+    (with-current-buffer (ewoc--buffer ewoc)
+      (setq stgit-index-node (cons ewoc index-node)
+            stgit-worktree-node (cons ewoc worktree-node)))))
 
 (defun stgit-get-position (&optional position)
   "Return position information at POSITION or point that can be restored later.


jpgrayson:
Thank you for this fix, @jrajahalme. Sorry for the delay in getting this reviewed and merged.

PR #639, opened 2026-06-21
fix(rebase): Run --exec commands also if the stack base does not change

### LINKED ISSUE #638: `stg rebase {base} --exec cmd` does nothing
Thank you for stgit. It's my favorite git tool.

I recently learned about the `rebase --exec` capability. I would like to execute a command for each currently applied patch without changing the base of the series. I tried to use the new `rebase --exec` for that. But it turns out that 

```
% stg rb {base} --exec "echo rebased"
info: Already based on `b958b246448bbba7ba971cb2cd7bc5b248262c04`
```

executes nothing. I was intuitively expecting the `--exec` to mechanically do the (trivial) rebase commit by commit instead of short-circuiting the rebase entirely. Similar to what git does:

```
 % git rebase --onto $(stg id {base}) $(stg id {base}) mybranch --exec "echo rebased"
Executing: echo rebased
rebased
Executing: echo rebased
rebased
Executing: echo rebased
rebased
Executing: echo rebased
rebased
Executing: echo rebased
rebased
Successfully rebased and updated refs/heads/mybranch.
```

Would it be reasonable to align the behavior with `git rebase --exec`?

mriemensberger:
`stg rebase --exec` is modeled after `git rebase --exec`. However, `stg rebase --exec cmd {base}` short-circuits the rebase entirely since the stack base does not change.  This diverges from `git rebase` behavior which execute commands in a similar scenario `git rebase --onto $(stg id {base}) $(stg id {base}) mybranch`.  Make stgit behave like git by removing the short-ciruiting if `--exec` is specified.

Fixes: #638

### DIFF
--- a/src/cmd/rebase.rs
+++ b/src/cmd/rebase.rs
@@ -121,6 +121,10 @@ fn run(matches: &ArgMatches) -> Result<()> {
     let allow_push_conflicts = argset::resolve_allow_push_conflicts(&config, matches);
     let committer_date_is_author_date = matches.get_flag("committer-date-is-author-date");
     let interactive = matches.get_flag("interactive");
+    let exec_cmds: Vec<&str> = matches
+        .get_many::<String>("exec")
+        .map(|vals| vals.map(String::as_str).collect())
+        .unwrap_or_default();
 
     let target_commit = if let Some(target_rev_spec) =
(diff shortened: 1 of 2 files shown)


jpgrayson:
Thank you for this contribution, @mriemensberger.

